PR #104, opened 2022-07-08
Add: Bounding Box concept on Page concept and manage their Relation to Documents

atraining:


konfuzio_sdk/data.py
```diff
@@ -72,6 +75,91 @@ def lose_weight(self):
...
+        """Get Document Page as PNG."""
+        if self.document.status[0] == 2 and (not is_file(self.image_path, raise_exception=False) or update):
+            png_content = get_page_image(self.id_)
+            with open(self.image_path, "wb") as f:
+                f.write(png_content)
+                self.image = Image.open(io.BytesIO(png_content))
+        elif is_file(self.image_path, raise_exception=False):
+            self.image = Image.open(self.image_path)
+        return self.image
+
```
samuel500, line 130: Methods to access the text and bboxes of the page could also be useful?

tests/test_data.py
```diff
@@ -151,7 +165,7 @@ def test_annotation_bbox(self):
             label_set=self.project.label_sets[0],
             spans=[span],
         )
-        span.bbox()
+        span.bbox
```
zypriafl, line 168: Why do we do this here?

konfuzio_sdk/data.py
```diff
@@ -1258,31 +1309,6 @@ def spans(self) -> List[Span]:
...
-        self.document = document
-        self.start_offset = start_offset
-        self.end_offset = end_offset
-        self.original_size = original_size
-        self.size = size
-        self.image = image
-        self.number = number
-
-
 class Document(Data):
```
zypriafl, line 1312: Document has a method get_bbox() however Span has method bbox(). We need to keep in mind that Span.bbox() returns are single Bbox instance. However Document.get_bbox() returns a Dictionary and does not use a Class yet.

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1295,10 +1295,10 @@ def Convert(tup, di):
 def annotation_to_dict(annotation: Annotation, include_pos: bool = False) -> dict:
     """Convert an annotation to a dictionary."""
     # Calculate area.
-    if annotation.spans[0].x0 is None or annotation.spans[0].y0 is None:
+    if annotation.spans[0].bbox().x0 is None or annotation.spans[0].bbox().y0 is None:
         raise NotImplementedError
     else:
-        area = annotation.spans[0].x0 * annotation.spans[0].y0
+        area = annotation.spans[0].bbox().x0 * annotation.spans[0].bbox().y0
```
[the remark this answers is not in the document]
atraining: The area of an Annotation itself does not provide much information. So we used the area that a Annotation creates across the page. But I am on your side, the variable name is misleading.

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1295,10 +1295,10 @@ def Convert(tup, di):
 def annotation_to_dict(annotation: Annotation, include_pos: bool = False) -> dict:
```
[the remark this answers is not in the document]
atraining: Let's say yes. The draft in information_extraction.py was created for Annotations that have only one Span. I added many todos due to that 
image

tests/test_data.py
atraining: good catch fixed, to test if we can calculate valid bboxes from a text offset. I added a comment in the code.

konfuzio_sdk/data.py
atraining: it's now part and document.bboxes returns real Bboxes that are validated according their dimensions x0, x1, y0, y1 and according if the document.text matches to the BoundingBox on a character level.

This will now make document.check_bboxes() obsolete.

PR #106, opened 2022-07-09
Add: Evaluation Class to slice and dice evaluation and add option to compare Documents less strict

atraining:


konfuzio_sdk/evaluate.py
```diff
@@ -104,11 +135,89 @@ def compare(doc_a, doc_b, only_use_correct=False, strict=True) -> pd.DataFrame:
...
+            query = f'label_set_id == {search.id_}'
+        else:
+            raise NotImplementedError
+        return self.data.query(query)
+
+    def tp(self, search=None) -> int:
+        """Return the True Positives of all Spans."""
+        return self.query(search=search)["true_positive"].sum()
+
+    def fp(self, search=None) -> int:
```
zypriafl, line 203: I think "false positive" function needs to consider the "_predicted" column too. For example if a Span has predicted Label "Firstname" and is a "false positive" it should be a "false postive" for the Label "Firstname". The same applies to "Label Set".

PR #108, opened 2022-07-14
Expanding Page functionalities

samuel500:


tests/test_api.py
```diff
@@ -207,7 +207,7 @@ def test_download_file_not_available(self):
     def test_get_annotations(self):
         """Download Annotations and the Text from API for a Document and check their offset alignment."""
         annotations = get_document_details(TEST_DOCUMENT_ID, project_id=TEST_PROJECT_ID)['annotations']
-        self.assertEqual(len(annotations), 22)
+        self.assertEqual(len(annotations), 21)
```
atraining, line 210: Why does the number of Annotations change in this commit?

atraining: if it's just due to changes in the online project it's fine. However, I am wondering what happened.

samuel500: Yes, it's a change in the online project. The same change was also already merged into master here: https://github.com/konfuzio-ai/konfuzio-sdk/pull/110/files

PR #110, opened 2022-07-15
Force data in offline mode

da-konfuzio:
Added: Option to force projects, documents, and related data in offline mode.

### DIFF
--- a/konfuzio_sdk/data.py
+++ b/konfuzio_sdk/data.py
@@ -2074,9 +2074,6 @@ class Document(Data):
                 raw_annotations = json.load(f)
 
             for raw_annotation in raw_annotations:
-                # todo should this filter?
-                if not raw_annotation['is_correct'] and raw_annotation['revised']:
-                    continue
                 raw_annotation['annotation_set_id'] = raw_annotation.pop('section')
                 raw_annotation['label_set_id'] = raw_annotation.pop('section_label_id')
                 _ = Annotation(document=self, id_=raw_annotation['id'], **raw_annotation)
@@ -2091,9 +2088,6 @@ class Document(Data):
                     raw_annotations = json.load(f)
 
                 for raw_annotation in raw_annotations:
-                    # todo should this filter?
-                    if not raw_annotation['is_correct'] and raw_annotation['revised']:
-                        continue
                     raw_annotation['annotation_set_id'] = raw_annotation.pop('section')
                     raw_annotation['label_set_id'] = raw_annotation.pop('section_label_id')
                     _ = Annotation(document=self, id_=raw_annotation['id'], **raw_annotation)
--- a/konfuzio_sdk/data.py
+++ b/konfuzio_sdk/data.py
@@ -42,6 +42,7 @@ class Data:
     id_ = None
     id_local = None
     session = _konfuzio_session()
+    _force_offline = False
 
     def __eq__(self, other) -> bool:
         """Compare any point of data with their ID, overwrite if needed."""
@@ -66,13 +67,19 @@ class Data:
     @property
     def is_online(self) -> Optional[int]:
         """Define if the Document is saved to the server."""
-        return self.id_ is not None
+        return (self.id_ is not None) and (not self._force_offline)
 
     # todo require to overwrite lose_weight via @abstractmethod
     def lose_weight(self):
         """Delete data of the instance."""
         self.session = None
         return self
+
+    def set_offline(self):
+        """Force data into offline mode."""
+        self._force_offline = True
+        if hasattr(self, '_update'):
+            self._update = False
 
 
 class Page(Data):
@@ -1559,7 +1566,7 @@ class Document(Data):
         """Return Annotation Sets of Documents."""
         if self._annotation_sets is not None:
             return self._annotation_sets
-        if self.id_ and not is_file(self.annotation_set_file_path, raise_exception=False):
+        if self.is_online and not is_file(self.annotation_set_file_path, raise_exception=False):
             self.download_document_details()
         if is_file(self.annotation_set_file_path, raise_exception=False):
             with open(self.annotation_set_file_path, "r") as f:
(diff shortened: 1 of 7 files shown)


konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -2629,10 +2629,14 @@ def feature_function(self, documents: List[Document], no_label_limit=None) -> Tu
...
+            if document.id_ is None:
                 # inference time todo reduce shuffled complexity
-                assert not label_annotations, 'Non online Documents have no human revised Annotations.'
-                raise NotImplementedError(f'{document} is not online, please us process_document_data function.')
+                assert (
+                    not label_annotations
+                ), "Documents that don't come from the server have no human revised Annotations."
+                raise NotImplementedError(
+                    f'{document} does not come from the server, please use process_document_data function.'
+                )
```
da-konfuzio, line 2639: Can now calculate features of documents as long as they have IDs, even if they are offline. The assumption is that if they have an ID, then the data came either from the API or from the DB.

tests/test_data.py
```diff
@@ -1193,7 +1193,7 @@ def test_make_sure_annotation_sets_are_downloaded_automatically(self):
         self.assertFalse(is_file(doc.annotation_set_file_path, raise_exception=False))
         self.assertEqual(None, doc._annotation_sets)
         self.assertTrue(doc.annotation_sets())
-        self.assertEqual(5, len(doc._annotation_sets))
+        self.assertEqual(4, len(doc._annotation_sets))
```
da-konfuzio, line 1196: As above: We are skipping this test but it's now passing after I run some django commands. Is there anything that will create the extra empty annotation sets or can I unskip the test?

konfuzio_sdk/data.py
```diff
@@ -2067,7 +2076,19 @@ def get_annotations(self) -> List[Annotation]:
...
+            self.annotation_sets()
             self._annotations = []
+            # We load the annotation file if it exists
+            if annotation_file_exists:
+                with open(self.annotation_file_path, 'r') as f:
+                    raw_annotations = json.load(f)
+
+                for raw_annotation in raw_annotations:
+                    if not raw_annotation['is_correct'] and raw_annotation['revised']:
+                        continue
```
atraining, line 2088: It should not filter. document.annotations() provides the filter to access Annotations.

konfuzio_sdk/data.py
```diff
@@ -66,14 +67,20 @@ def __deepcopy__(self, memodict):
...
     # todo require to overwrite lose_weight via @abstractmethod
     def lose_weight(self):
         """Delete data of the instance."""
         self.session = None
         return self
 
+    def set_offline(self):
+        """Force data into offline mode."""
+        self._force_offline = True
+        if hasattr(self, '_update'):
```
atraining, line 81: Please work without hasattr. In the past it has created a lot of undefined situation, see informaiton_extraction.py - Either Make sure every class has a _update or remove the _update from all.

tests/test_data.py
```diff
@@ -1656,6 +1656,45 @@ def tearDownClass(cls) -> None:
...
+        """Test if Pages are downloaded automatically."""
+        prj = Project(id_=TEST_PROJECT_ID, project_folder='another33')
+        doc = prj.get_document_by_id(TEST_DOCUMENT_ID)
+        doc.set_offline()
+        self.assertFalse(is_file(doc.pages_file_path, raise_exception=False))
+        self.assertEqual([], doc._pages)
+        self.assertFalse(doc.pages())
+        self.assertFalse(is_file(doc.pages_file_path, raise_exception=False))
+        prj.delete()
+
```
atraining, line 1696: Add tests if id_ is given and instance is set offline for Label, Label Set, Annotation Set, Annotation, ... any Class in data.py

tests/test_data.py
```diff
@@ -1656,6 +1656,45 @@ def tearDownClass(cls) -> None:
...
+    def test_make_sure_annotations_are_not_downloaded_automatically(self):
+        """Test if Annotations are downloaded automatically."""
+        prj = Project(id_=TEST_PROJECT_ID, project_folder='another')
+        doc = prj.get_document_by_id(TEST_DOCUMENT_ID)
+        doc.set_offline()
+        self.assertFalse(is_file(doc.annotation_file_path, raise_exception=False))
+        self.assertEqual(None, doc._annotations)
+        self.assertFalse(doc.annotations())
+        self.assertEqual(0, len(doc._annotations))
```
atraining, line 1670: What do you think to add one test, where you verify to load annotations from Project with ID TEST_PROJECT_ID and from path OFFLINE_PROJECT and initialize Annotations TEST_DOCUMENT_ID  even this project is set_offline.

konfuzio_sdk/data.py
```diff
@@ -66,14 +67,20 @@ def __deepcopy__(self, memodict):
     @property
     def is_online(self) -> Optional[int]:
         """Define if the Document is saved to the server."""
-        return self.id_ is not None
+        return (self.id_ is not None) and (not self._force_offline)
```
atraining, line 70: I am not quite sure if there are really 3 states: 1. Is online = Data is available one public host, is local = data is serialized to disk, is memory = Project is initialized in memory and might be saved to disk or online, something to clarify with mexicat.

mexicat: I think these three states are correct but in the last case (Project is initialized in memory) from the server's perspective we don't need to save anything online because everything is saved directly on the database by the server based on the results of extraction/evaluation/training.

I'm not sure if there are cases in the SDK where something is initialized in memory and needs to be saved online - again, in the server this is not needed.

konfuzio_sdk/data.py
```diff
@@ -2043,22 +2056,25 @@ def evaluate_regex(self, regex, label: Label, annotations: List['Annotation'] =
...
 
             self._annotations = []  # clean Annotations to not create duplicates
+            # We read the annotation file that we just downloaded
             with open(self.annotation_file_path, 'r') as f:
                 raw_annotations = json.load(f)
 
             for raw_annotation in raw_annotations:
+                # todo should this filter?
                 if not raw_annotation['is_correct'] and raw_annotation['revised']:
                     continue
```
da-konfuzio, line 2079: If it shouldn't filter when the annotations are None, then it shouldn't filter here either?

konfuzio_sdk/trainer/information_extraction.py
zypriafl: Can you add a comment why an ID is needed to calculate features?

PR #111, opened 2022-07-19
Add: Inside–outside–beginning (tagging) when Annotations in one Document overlap

samuel500:
bug fix in `convert_to_bio_scheme` which seems to take care of the duplicates.


konfuzio_sdk/samples.py
```diff
@@ -99,3 +99,92 @@ def __init__(self):
             label_set=label_set,
             spans=[Span(start_offset=0, end_offset=7)],
         )
+
+
+class LocalViewProject(Project):
```
atraining, line 104: can we not add this Document to LocalTextProject, so we have the complexity in one class?

konfuzio_sdk/samples.py
```diff
@@ -99,3 +99,92 @@ def __init__(self):
...
+uncertain
+"""
+        document = Document(project=self, category=category, text=doc_3_text, dataset_status=2)
+        label_set = LabelSet(id_=2, project=self, categories=[category])
+        label = Label(id_=1, text='LabelName', project=self, label_sets=[label_set])
+        label_date = Label(id_=2, text='date', project=self, label_sets=[label_set], has_multiple_top_candidates=False)
+        label_3 = Label(id_=3, text='LabelName 3', project=self, label_sets=[label_set])
+        label_4 = Label(id_=4, text='LabelName 4', project=self, label_sets=[label_set], threshold=0.1)
+
+        # to be displayed in smart view: #16, #18, #19
```
atraining, line 126: we might want to add one Annotation that is is_correct=False, is_revised=True, confidence = 1 and has Span 37 to 49 - > it should not be displayed

konfuzio_sdk/data.py
```diff
@@ -1722,6 +1722,43 @@ def annotations(
 
         return sorted(annotations)
 
+    def view_annotations(self) -> List[Annotation]:
+        """Get all annotations viewable in the smart-view. Filter out all others."""
+        if self.category is None:
```
atraining, line 1727: it's a good check but perhaps it should be added to get_annotations directly?

konfuzio_sdk/samples.py
```diff
@@ -99,3 +99,92 @@ def __init__(self):
...
+            spans=[Span(start_offset=0, end_offset=5)],
+        )
+
+        _ = Annotation(
+            id_=16,
+            document=document,
+            is_correct=True,
+            label=label,
+            label_set=label_set,
+            spans=[Span(start_offset=0, end_offset=5)],
```
atraining, line 143: I did not find the situation where two Annotations are is_correct = False, is_revised=False but one Annotation has higher confidence than the other and starts later

example

Annotation, confidence = 0.9, Span(2,5)
Annotation, confidence = 0.99, Span(3,4)

konfuzio_sdk/samples.py
samuel500: A number of tests in test_evaluate.py started failing when I added it to LocalTextProject. Maybe I could try to resolve those instead, but I think it might be better to have these different tests more disentangled?

konfuzio_sdk/samples.py
samuel500: Is this because it would mean that the user marked the annotation as incorrect?

konfuzio_sdk/data.py
samuel500: Maybe? I just copied the first 3 lines of  `def annotations` here. If we add the check to get_annotations, I guess we also would be able to remove the check from `def annotations`?

konfuzio_sdk/samples.py
samuel500: Fixed by Davide

PR #114, opened 2022-07-22
Fix: Reduce the runtime of ListTokenizers by remove ListTokenizer duplicates

samuel500:


konfuzio_sdk/tokenizer/regex.py
```diff
@@ -22,6 +23,17 @@ def __repr__(self):
         """Return string representation of the class."""
         return f"{self.__class__.__name__}: {repr(self.regex)}"
 
+    def __hash__(self):
+        """Get unique hash for RegexTokenizer."""
+        return hash(repr(self))
```
atraining, line 28: def __hash__(self):
   return hash(repr(self.regex))

konfuzio_sdk/tokenizer/regex.py
```diff
@@ -22,6 +23,17 @@ def __repr__(self):
...
+
+    def __eq__(self, other) -> bool:
+        """Compare RegexTokenizer with another Tokenizer."""
+        if not isinstance(other, RegexTokenizer):
+            return False
+        else:
+            return self.regex == other.regex
```
atraining, line 35: return self== other

samuel500: This gives me a recursion error. And the default __eq__ doesn't seem to use __hash__. This seems to work though: `return hash(self) == hash(other)`

PR #118, opened 2022-08-01
Allow setting the strict parameter when initializing Evaluation

mexicat:
Needed for non-strict evaluation from the server.

### DIFF
--- a/konfuzio_sdk/evaluate.py
+++ b/konfuzio_sdk/evaluate.py
@@ -217,14 +217,15 @@ class Evaluation:
 
     from konfuzio_sdk.data import Document
 
-    def __init__(self, documents: List[Tuple[Document, Document]]):
+    def __init__(self, documents: List[Tuple[Document, Document]], strict: bool = True):
         """
         Relate to the two document instances.
 
         :param documents: A list of tuple Documents that should be compared.
+        :param strict: A boolean passed to the `compare` function.
         """
         self.documents = documents
-        self.strict = True
+        self.strict = strict
         self.only_use_correct = True
         self.data = None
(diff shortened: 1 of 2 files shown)


konfuzio_sdk/evaluate.py
```diff
@@ -217,14 +217,15 @@ class Evaluation:
 
     from konfuzio_sdk.data import Document
 
-    def __init__(self, documents: List[Tuple[Document, Document]]):
+    def __init__(self, documents: List[Tuple[Document, Document]], strict: bool = True):
         """
         Relate to the two document instances.
 
         :param documents: A list of tuple Documents that should be compared.
+        :param strict: A boolean passed to the `compare` function.
```
atraining, line 225: `By default the evaluation is strict. A strict evaluation compares the exact start and end offset of a Span. If the evaluation is not strict the predicted text just needs to be within the annotated text by the human.`

konfuzio_sdk/evaluate.py
```diff
@@ -217,14 +217,15 @@ class Evaluation:
...
         """
         self.documents = documents
-        self.strict = True
```
atraining, line 227: Please add one test for strict = False using the `__init__`.

konfuzio_sdk/evaluate.py
mexicat: Do I need to add this text to the docstring? I think this should be explained in the `compare` function instead.

konfuzio_sdk/evaluate.py
mexicat: Done.

PR #120, opened 2022-08-08
Added: check negative coordinates for bboxes

da-konfuzio:


konfuzio_sdk/data.py
```diff
@@ -112,10 +112,10 @@ def __init__(
         if self.index is None:
             logger.error(f'Page index is None of {self} in {self.document}.')
             check_page = False
-        if self.height is None:
+        if self.height is None:  # todo why do we allow pages with height<=0?
```
atraining, line 115: can u solve this todo

atraining: @da-konfuzio was this intended, otherwise can we remove the bbox

image

da-konfuzio: Hi, yes it was intended, see
https://gitlab.com/konfuzio/objectives/-/issues/9269#note_1052318423
https://gitlab.com/konfuzio/objectives/-/issues/9269#note_1053619707

PR #122, opened 2022-08-28
Changed: Checks that disallow zero width/height bboxes are now optional

da-konfuzio:


konfuzio_sdk/data.py
```diff
@@ -227,16 +227,15 @@ def __eq__(self, other: 'Bbox') -> bool:
...
+        Validate contained data.
 
+        We allow zero width/height in bboxes for compatibility reasons since some OCR engines can sometimes return such
+        kinds of character level bboxes.
+        """
         if self.x0 > self.x1:
             raise ValueError(f'{self} has negative width in {self.page}.')
 
-        if self.y0 == self.y1:
-            raise ValueError(f'{self} has no height in {self.page}.')
```
atraining, line 238: why don't you keep an error log? how will null height bounding boxes impact feature generation. perhaps this is something to document + link to the documentation form the log error.

tests/test_data.py
```diff
@@ -580,38 +578,23 @@ def test_document_check_bbox_coordinates(self):
         _ = Page(id_=1, number=1, original_size=(595.2, 841.68), document=document, start_offset=0, end_offset=1)
         self.assertTrue(document.check_bbox())
 
-    def test_document_check_bbox_invalid_x_coordinates(self):
```
atraining, line 583: can you make sure we have tests which test for invalid documents / coordinates, I see that you overwrote this test but did not add a new one

tests/test_data.py
```diff
@@ -580,38 +578,23 @@ def test_document_check_bbox_coordinates(self):
...
             '0': {'x0': 1, 'x1': 1, 'y0': 0, 'y1': 1, 'top': 10, 'bottom': 11, 'page_number': 1, 'text': 'h'}
         }
         document = Document(project=self.project, category=self.category, text='h', bbox=document_bbox)
         _ = Page(id_=1, number=1, original_size=(595.2, 841.68), document=document, start_offset=0, end_offset=1)
-        with pytest.raises(ValueError) as e:
-            self.assertFalse(document.check_bbox())
-            assert 'has negative or no height' in str(e)
+        document.bboxes
 
-    def test_document_check_bbox_invalid_y_coordinates(self):
```
atraining, line 594: see above

konfuzio_sdk/data.py
```diff
@@ -227,16 +227,15 @@ def __eq__(self, other: 'Bbox') -> bool:
     def _valid(
         self,
     ):
-        """Validate contained data."""
-        if self.x0 == self.x1:
-            raise ValueError(f'{self} has no width in {self.page}.')
```
atraining, line 232: suggestion: keep an error log or add strict=False as _valid() argument?

 how will null height bounding boxes impact feature generation. perhaps this is something to document + link to the documentation form the log error.

tests/test_data.py
da-konfuzio: Ok, I'll review whether we already have them (I think so for negative coordinates and dimensions) and add what's missing.

konfuzio_sdk/data.py
da-konfuzio: Done, but we need to have a page on dev.konfuzio.com about general feature function documentation where we also include these impacts from zero dimensions in bboxes, which depends on internal ticket 9523

konfuzio_sdk/data.py
```diff
@@ -1436,6 +1458,7 @@ def __init__(
         category: Category = None,
         text: str = None,
         bbox: dict = None,
+        strict_bbox_validation: bool = False,
```
da-konfuzio, line 1461: Just an initial suggestion: now we have an additional argument in the Document's constructor, not sure if there are better ways because bboxes don't get created by the user (just a dict is provided and then the Document creates the object internally) so the information about strict_validation of bboxes has to be propagated from the Document object itself.

I don't like it so much but it still seems valuable to have this strict option for bbox dimensions.

tests/test_data.py
```diff
@@ -484,13 +484,32 @@ def test_span_reference_to_annotation(self):
...
+    def test_get_span_bbox_with_characters_without_height_strict_validation(self):
+        """
+        Test get the bbox of a Span where the characters do not have height (OCR problem).
+
+        With strict validation specified, we don't allow such bboxes.
+        """
+        document_bbox = {'1': {'text': 'e', 'x0': 0, 'x1': 1, 'y0': 1, 'y1': 1, 'page_number': 1}}
+        document = Document(
+            project=self.project, category=self.category, text='hello', bbox=document_bbox, strict_bbox_validation=True
+        )
```
da-konfuzio, line 509: As I said, the user just provides a dict so the Document object needs to know whether to internally check bbox size in a strict way or not.

konfuzio_sdk/data.py
atraining: Added a task to the ticket. Perhaps you can adjust the docstring of "_valid" to reflect the thoughts of those strict or non strict options to validate a bbox.

PR #133, opened 2022-09-27
Added: Methods to retrieve and delete online annotations

iftwigs:
It's now possible to:
- retrieve an annotation by its ID by using `Document.get_annotation_by_id()` (see [Python API docs](https://dev.konfuzio.com/sdk/sourcecode.html#document))
- delete an online annotation by using `Annotation.delete()` (see [Python API docs](https://dev.konfuzio.com/sdk/sourcecode.html#annotation))


tests/test_data.py
```diff
@@ -111,6 +111,48 @@ def test_load_image_in_memory(self):
...
+        label = self.project.labels[8]
+        annotation = Annotation(
+            document=doc,
+            spans=[Span(start_offset=1603, end_offset=1605)],
+            label=label,
+            label_set=label.label_sets[0],
+            accuracy=1.0,
+            is_correct=True,
+        )
+        annotation.save()
```
da-konfuzio, line 150: It's redundant to add this annotation online, since the testcase is for offline annotations

tests/test_data.py
```diff
@@ -111,6 +111,48 @@ def test_load_image_in_memory(self):
...
+
+        assert Span(start_offset=1590, end_offset=1602) in doc.spans()
+
+    def test_delete_annotation_online(self):
+        """Test to delete an Annotation from the document online."""
+        doc = self.project.get_document_by_id(TEST_DOCUMENT_ID)
+        annot = [x for x in doc.get_annotations() if x.start_offset == 1590 and x.end_offset == 1602]
+        annot[0].delete()
+
+        assert annot[0] not in doc.get_annotations()
```
da-konfuzio, line 136: Can you please add the following:

Before this step, what we need to do to really check that the annotation is deleted online, is to sync the local Document object with the online version, so that if somehow the annotation is still there, it's downloaded again. This wouldn't be detected by the single assertion because Annotation.delete() is deleting the local annotation.

tests/test_data.py
```diff
@@ -111,6 +111,48 @@ def test_load_image_in_memory(self):
...
+            document=doc,
+            spans=[Span(start_offset=1590, end_offset=1602)],
+            label=label,
+            label_set=label.label_sets[0],
+            accuracy=1.0,
+            is_correct=True,
+        )
+        annotation.save()
+
+        assert Span(start_offset=1590, end_offset=1602) in doc.spans()
```
da-konfuzio, line 128: Can you please add the following:

Before this step, what we need to do to really check that the annotation is created online, is to sync the local Document object with the online version, so that if somehow the annotation is not added online, we don't download it. This wouldn't be detected by the single assertion because Annotation.save() is creating the local annotation.

You can do this with doc.update()

konfuzio_sdk/data.py
```diff
@@ -1478,11 +1479,34 @@ def regex(self):
...
+                if annotation == self:
+                    with open(
+                        'data_{}/documents/{}/annotations.json5'.format(self.document.project.id_, self.document.id_),
+                        'r',
+                    ) as f:
+                        annots = json.load(f)
+                    for el in annots:
+                        for bbox in el['bboxes']:
+                            if bbox['start_offset'] == self.start_offset and bbox['end_offset'] == self.end_offset:
+                                el['bboxes'].remove(bbox)
```
da-konfuzio, line 1509: Here the element that gets deleted is not part of the object attributes. I think the old logic:
`del self.document._annotations[index]`

Was correct, can you please add it back in?

konfuzio_sdk/data.py
```diff
@@ -1478,11 +1479,45 @@ def regex(self):
...
-            if annotation == self:
-                del self.document._annotations[index]
+    def delete(self, delete_online=True) -> None:
+        """Delete Annotation."""
+        if self.document.is_online and delete_online:
+            for index, annotation in enumerate(self.document._annotations):
+                if annotation == self:
+                    delete_document_annotation(self.document.id_, self.id_, self.document.project.id_)
+                    with open(
+                        'data_{}/documents/{}/annotations.json5'.format(self.document.project.id_, self.document.id_),
```
atraining, line 1489: please use the Python API to access the path

konfuzio_sdk/data.py
```diff
@@ -1478,11 +1479,45 @@ def regex(self):
...
+                        'r',
+                    ) as f:
+                        annots = json.load(f)
+                    for el in annots:
+                        for bbox in el['bboxes']:
```
atraining, line 1494: can we use the data layer?

konfuzio_sdk/data.py
```diff
@@ -1478,11 +1479,45 @@ def regex(self):
...
+                            if bbox['start_offset'] == self.start_offset and bbox['end_offset'] == self.end_offset:
+                                el['bboxes'].remove(bbox)
+                    with open(
+                        'data_{}/documents/{}/annotations.json5'.format(self.document.project.id_, self.document.id_),
```
atraining, line 1516: see above - please use the python api

konfuzio_sdk/data.py
```diff
@@ -1478,11 +1479,45 @@ def regex(self):
...
-    def delete(self) -> None:
-        """Delete Annotation online."""
-        for index, annotation in enumerate(self.document._annotations):
-            if annotation == self:
-                del self.document._annotations[index]
+    def delete(self, delete_online=True) -> None:
+        """Delete Annotation."""
+        if self.document.is_online and delete_online:
+            for index, annotation in enumerate(self.document._annotations):
+                if annotation == self:
```
atraining, line 1486: why do we need to iterade? why did you not implement self.document.get_annotation_by_id

konfuzio_sdk/data.py
```diff
@@ -1478,11 +1479,45 @@ def regex(self):
...
+                    delete_document_annotation(self.document.id_, self.id_, self.document.project.id_)
+                    with open(
+                        'data_{}/documents/{}/annotations.json5'.format(self.document.project.id_, self.document.id_),
+                        'r',
+                    ) as f:
+                        annots = json.load(f)
+                    for el in annots:
```
atraining, line 1493: please use sounding variable names - el is quite hard to get isn't it annotation in annotations

konfuzio_sdk/data.py
```diff
@@ -1478,11 +1479,45 @@ def regex(self):
...
 
-    def delete(self) -> None:
-        """Delete Annotation online."""
-        for index, annotation in enumerate(self.document._annotations):
-            if annotation == self:
-                del self.document._annotations[index]
+    def delete(self, delete_online=True) -> None:
+        """Delete Annotation."""
+        if self.document.is_online and delete_online:
+            for index, annotation in enumerate(self.document._annotations):
```
atraining, line 1485: why do you iterate _annotations? - why do we delete on an Annotation (the method) level but then iterate on a document level

konfuzio_sdk/data.py
iftwigs: I modified the existent delete() method which included the iterating, there was no other reason for that, will fix

iftwigs: also I can't seem to find self.document.get_annotation_by_id, at least not in the last version present in the repo.

atraining: good point to create and so this functionality can be reused

I mean create a way to get an annotation by id - perhaps it's on the Project level or Document Level

or has to be created - please discuss with Davide

iftwigs: oh sure, makes sense. we'll discuss it with Davide then

konfuzio_sdk/data.py
iftwigs: to my understanding, this was necessary to locate the index of the annotation for deleting it further. I have rewritten the logic into a list comprehension but I think there's no way to avoid it at all until I've implemented the method you've mentioned above.

konfuzio_sdk/data.py
iftwigs: what do you mean by the data layer?

konfuzio_sdk/data.py
```diff
@@ -1478,11 +1479,30 @@ def regex(self):
...
         return self.label.combined_tokens(categories=[self.document.category])
 
-    def delete(self) -> None:
-        """Delete Annotation online."""
-        for index, annotation in enumerate(self.document._annotations):
-            if annotation == self:
-                del self.document._annotations[index]
+    def delete(self, delete_online=True) -> None:
+        """Delete Annotation."""
+        index = [x for x, y in enumerate(self.document._annotations) if y == self][0]
```
da-konfuzio, line 1484: Should be using `y is self`

konfuzio_sdk/data.py
```diff
@@ -1478,11 +1479,30 @@ def regex(self):
...
+        else:
+            with open(self.document.annotation_file_path, 'r') as f:
+                annots = json.load(f)
+            for annot in annots:
+                for bbox in annot['bboxes']:
+                    if bbox['start_offset'] == self.start_offset and bbox['end_offset'] == self.end_offset:
+                        annot['bboxes'].remove(bbox)
+            with open(self.document.annotation_file_path, 'w') as f:
+                json.dump(annots, f)
+            del self.document._annotations[index]
```
da-konfuzio, line 1505: This duplicated code fragment can be unified by unindenting it and executing it once outside of the initial `if self.document.is_online and delete_online:`

tests/test_data.py
```diff
@@ -111,6 +111,46 @@ def test_load_image_in_memory(self):
             image = page.get_image()
             assert type(image) is PngImageFile
 
+    def test_create_annotation(self):
+        """Test to add an Annotation to the document."""
+        doc = self.project.get_document_by_id(TEST_DOCUMENT_ID)
+        label = self.project.labels[8]
```
da-konfuzio, line 117: use `get_label_by_name` for readability and consistency (maybe the index of the label won't be 8 in the future)

tests/test_data.py
```diff
@@ -111,6 +111,46 @@ def test_load_image_in_memory(self):
...
+        annotation = Annotation(
+            document=doc,
+            spans=[Span(start_offset=1590, end_offset=1602)],
+            label=label,
+            label_set=label.label_sets[0],
```
da-konfuzio, line 122: use `get_label_set_by_name` for readability and consistency (maybe the index of the label won't be 0 in the future)

tests/test_data.py
```diff
@@ -111,6 +111,46 @@ def test_load_image_in_memory(self):
...
+        annotation = Annotation(
+            document=doc,
+            spans=[Span(start_offset=1603, end_offset=1605)],
+            label=label,
+            label_set=label.label_sets[0],
+            accuracy=1.0,
+            is_correct=True,
+        )
+        annotation.delete(delete_online=False)
+        doc = self.project.get_document_by_id(TEST_DOCUMENT_ID)
```
da-konfuzio, line 151: not necessary to get the document locally again

tests/test_data.py
```diff
@@ -111,6 +111,46 @@ def test_load_image_in_memory(self):
...
+        assert annotation not in doc.get_annotations()
```
da-konfuzio, line 152: then I would add another check to see that it wasn't deleted online as well

thus `doc.update()`

and then `assert annotation in doc.get_annotations()`

konfuzio_sdk/data.py
da-konfuzio: It's actually not necessary to iterate, we can do

`self.document._annotations.remove(self)`

instead of:

`del self.document._annotations[index]`

because `list.remove` uses `is` internally

:wink:

tests/example_project_data/documents/44823/bio_scheme.txt
```diff
@@ -0,0 +1,384 @@
+x02 O
```
da-konfuzio, line 1: Can you double check that this bio_scheme is not included in the branch?

tests/example_project_data/regex/EMPTY_LABEL.json5
```diff
@@ -0,0 +1 @@
+[]
```
da-konfuzio, line 1: As with the bio_scheme, can you check that this json file is not included?

tests/example_project_data/documents/44823/bio_scheme.txt
iftwigs: yes sorry, this must have leaked from another branch here

PR #135, opened 2022-09-28
Added: Tokenizer Information to Spans

samuel500:


konfuzio_sdk/tokenizer/regex.py
```diff
@@ -87,6 +92,25 @@ def tokenize(self, document: Document) -> Document:
 
         return document
 
+    def span_match(self, span: 'Span'):
+        """Check if offset is detected by Tokenizer."""
+        if self in span.regex_matching:
+            return True
+        else:
+            slice_start = max(0, span.start_offset - 25)
+            slice_end = min(span.end_offset + 5, len(span.annotation.document.text))
```
da-konfuzio, line 101: What do these numbers (-25 and +5) stand for?

konfuzio_sdk/tokenizer/regex.py
```diff
@@ -87,6 +92,25 @@ def tokenize(self, document: Document) -> Document:
 
         return document
 
+    def span_match(self, span: 'Span'):
```
da-konfuzio, line 95: Let's say I'm working on Server code and I want to get a list of all Annotations (given some Label) that can't be found by the associated tokenizer.

I would have to code something like:
```python
cant_be_found_annotations: Dict[Label, List[Annotation] = {}
tokenizers: Dict[Label, Tokenizer] = train_tokenizers()  # whatever trains the tokenizers

for label in labels:
    cant_be_found_annotations[label] = []
    label_tokenizer = tokenizers[label]
    for annotation in label.annotations(categories=[category])
        match = all([label_tokenizer.span_match(span) for span in annotation.spans])
        if not match:
            cant_be_found_annotations[label].append(annotation)
```

This is fine. Perhaps it would be even better if there was a single, tested method for getting these. Such as
```python
Label.annotations_not_by_tokenizer(categories: List[Category], tokenizer: Tokenizer) -> List[Annotation]
```

Please ignore this comment if Server team already expressed a preference for checking annotations one by one on their side.

atraining: @da-konfuzio @samuel500 please add this to the documentation.

be aware the code is not also used by the Server but other SDK developers who want to find Spans that are not found by their Tokenizer

so rather implement the method and test it

konfuzio_sdk/data.py
```diff
@@ -874,6 +874,16 @@ def regex(self, categories: List[Category], update=False) -> List:
         categories_ids = [category.id_ for category in categories]
         return {k: v for k, v in self._regex.items() if k in categories_ids}
 
+    def spans_not_found_by_tokenizer(self, tokenizer, categories: List[Category], use_correct=False) -> List['Span']:
```
da-konfuzio, line 877: Can you add a small boilerplate (probably less than 10 lines of code) as an example for how to use this, under https://dev.konfuzio.com/sdk/examples/examples.html ?

PR #138, opened 2022-10-07
Added: Evaluation method for all AI training stages

samuel500:


konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -2867,174 +2867,86 @@ def evaluate_full(self, strict: bool = True) -> Evaluation:
...
 
-        self.evaluation = Evaluation(eval_list, strict=strict)
+        self.full_evaluation = Evaluation(eval_list, strict=strict)
 
-        return self.evaluation
+        return self.full_evaluation
 
-    def data_quality(self, strict: bool = True) -> Evaluation:
-        """Evaluate the full pipeline on the pipeline's Training Documents."""
+    def evaluate_tokenizer(self) -> Evaluation:
```
da-konfuzio, line 2874: 1. Since we have the `data_quality` method corresponding to `evaluate_full`, we should probably have the corresponding ways to evaluate on the training set for each of the extraction stages. Perhaps a boolean flag `evaluate_on_training_docs` can be enough.

2. Given the level of detail in these new evaluation results, it would be good to have a single code example to showcase all the evaluation stages under https://dev.konfuzio.com/sdk/examples/examples.html

da-konfuzio:
@samuel500 after merging `9574-loading-model-pickle` and `9549-add-tokenizer-information` into `9511-single-extraction-ai`, there is a conflict in the new `missing_spans` method when merging this branch as well. Can you solve it?

PR #150, opened 2022-11-09
Added: Rule based logic for multi-file Document splitting

iftwigs:
This new feature is a rule based, Context-aware file splitting logic for splitting files consisting of more than one Document. It allows for either pointing at possible split points (Pages that could be start of a new Document) in the Document or for returning a set of sub-Documents.

Find the following documentation:

- [Code Examples: Splitting for multi-file Documents, step-to-step guide](https://dev.konfuzio.com/sdk/examples/examples.html#splitting-for-multi-file-documents-step-by-step-guide)
- [Code Examples: out-of-the-box usage of File Splitting class](https://dev.konfuzio.com/sdk/examples/examples.html#split-multi-file-document-into-separate-files-without-training-a-model)
- [Code Examples: Evaluation of File Splitting class](https://dev.konfuzio.com/sdk/examples/examples.html#filesplittingevaluation-class)


docs/sdk/examples/file_splitting.md
```diff
@@ -0,0 +1,64 @@
+## Split Multi-File Documents into Separate Files
+
+Let's see how to use the `konfuzio_sdk` to create a pipeline for automatically splitting documents consisting of 
+several files.
+
+```python
+from konfuzio_sdk.data import Project
+from konfuzio_sdk.trainer import file_splitting
```
da-konfuzio, line 8: FYI the pattern we use in ExtractionAI and CategorizationAI is the equivalent of

```python
from konfuzio_sdk.trainer.file_splitting import FileSplittingModel, FileSplittingAI
```

docs/sdk/examples/file_splitting.md
```diff
@@ -0,0 +1,64 @@
...
+
+# train a file-splitting model on Documents from your Project 
+
+project = Project(id_=YOUR_PROJECT_ID)
+train_documents = project.documents
+test_documents = project.test_documents
```
da-konfuzio, line 14: To make it in line with the ExtractionAI template and the CategorizationAI template, I would like to do similarly in the SplittingAI template:

```python
file_splitting_model = FileSplittingModel(args)  # whatever args necessary
file_splitting_model.documents = project.documents
file_splitting_model.test_documents = project.test_documents
```

docs/sdk/examples/file_splitting.md
```diff
@@ -0,0 +1,64 @@
...
+
+train_img_data, train_txt_data, test_img_data, test_txt_data, train_labels, test_labels, txt_input_shape = file_splitting.FileSplittingModel.prepare_visual_textual_data(
+                                                                                                                                                                        train_documents,
+                                                                                                                                                                        test_documents,
+                                                                                                                                                                        bert_model,
+                                                                                                                                                                        bert_tokenizer)
+
+# initialize a fusion model
+
+model = file_splitting.FileSplittingModel.init_model(txt_input_shape)
```
da-konfuzio, line 30: Similarly as above, to keep the template looking like the one of the existing ExtractionAI, we can do the following.

We show the SDK concepts to the user, and do data processing within the methods of FileSplittingModel and FileSplittingAI. In this case, we can modify `FileSplittingModel.init_model`, such that:

1. `bert_model, bert_tokenizer = file_splitting.FileSplittingModel.init_bert()` happens inside of `FileSplittingModel.init_model`, because the bert_model and bert_tokenizer are not SDK concepts. We can still make them accessible as `FileSplittingModel.bert_model` and `FileSplittingModel.bert_tokenizer`
2. The same for `FileSplittingModel.prepare_visual_textual_data`, it should be possible to do this step within `FileSplittingModel.init_model` since all required variables are known to the class (and become publicly accessible)

So we are left with:

```python
file_splitting_model.init_model(
    documents=file_splitting_model.documents, test_documents=file_splitting_model.test_documents
)
```

What you currently call `model` is the page classifier, right? Then we can provide it as `file_splitting_model.page_classifier`

docs/sdk/examples/file_splitting.md
```diff
@@ -0,0 +1,64 @@
...
+
+# train a model on the data from the project and save the result
+
+model.fit([train_img_data, train_txt_data], train_labels, epochs=10, verbose=1)
+model.save(project.model_folder + '/fusion.h5')
```
da-konfuzio, line 35: wrap into `file_splitting_model.save`

docs/sdk/examples/file_splitting.md
```diff
@@ -0,0 +1,64 @@
...
+
+# evaluate the model
+
+loss, acc = model.evaluate([test_img_data, test_txt_data], test_labels, verbose=0)
+print('Accuracy: {}'.format(acc * 100))
+precision, recall, f1 = file_splitting.FileSplittingModel.calculate_metrics(model, 
+                                                                            test_img_data, 
+                                                                            test_txt_data,
+                                                                            test_labels)
+print('\n Precision: {} \n Recall: {} \n F1-score: {}'.format(precision, recall, f1))
```
da-konfuzio, line 45: Let's only show the metrics calculated for proper file splitting, otherwise we are counting all the `not_first_pages` classified correctly in the `acc` which artificially increases the score (i.e. true negatives)

```python
precision, recall, f1 = file_splitting_model.calculate_metrics()
```

docs/sdk/examples/file_splitting.md
```diff
@@ -0,0 +1,64 @@
...
+prediction = splitting_ai.propose_split_documents(test_documents[0])
+
+# iterate over the test documents to predict if any of them need to be split
+
+split_documents = {}
+
+for doc in test_documents:
+    prediction = splitting_ai.propose_split_documents(doc)
+    if len(prediction) > 1:
+        split_documents[doc.id_] = prediction
```
da-konfuzio, line 63: Not necessary

docs/sdk/examples/file_splitting/file_splitting_without_model.md
```diff
@@ -0,0 +1,48 @@
...
+splitting_ai = SplittingAI(project_id=project.id_)
+
+# process the test Document because it needs to be tokenized by the same Tokenizer as the Documents used by 
+# ContextAwareFileSplittingModel for gathering Spans unique for the first Pages. deepcopy is used to remove all previous
+# Annotations.
+test_document = file_splitting_model.tokenizer.tokenize(deepcopy(test_document))
+
+# propose a list of sub-Documents. If a test Document consists of a single file, return will consist of a list with the 
+# same Document.
+split_documents = splitting_ai.propose_split_documents(test_document)
```
da-konfuzio, line 47: The tokenization step should go within the `predict` method of the fallback class :wink:

tests/trainer/test_file_splitting.py
```diff
@@ -0,0 +1,96 @@
...
+            if category.id_ == 3:
+                self.file_splitting_model.predict(page)
+                if page.number == 1:
+                    assert intersections == [{'I like bread.'}]
+                    assert hasattr(page, 'is_first_page')
+                if page.number in (2, 4):
+                    assert intersections == []
+                if page.number in (3, 5):
+                    assert intersections == [{'Morning,'}]
+                    assert hasattr(page, 'is_first_page')
```
da-konfuzio, line 79: Should this be indented one level deeper? I see the `category` variable being used outside of the for loop

tests/trainer/test_file_splitting.py
```diff
@@ -0,0 +1,96 @@
...
+    def test_suggest_first_pages(self):
+        """Test SplittingAI's suggesting first Pages."""
+        splitting_ai = SplittingAI(self.file_splitting_model)
+        pred = splitting_ai.propose_split_documents(self.test_document, return_pages=True)
+        for page in pred.pages():
+            print(hasattr(page, 'is_first_page'))
+            if page.number in (1, 3, 5):
+                assert hasattr(page, 'is_first_page')
+            else:
+                assert not hasattr(page, 'is_first_page')
```
da-konfuzio, line 96: We can directly check the truth value of `page.is_first_page`, right? We no longer need hasattr :wink:

tests/test_data.py
```diff
@@ -2173,7 +2200,7 @@ def test_make_sure_pages_are_not_downloaded_automatically(self):
     def test_view_annotations(self):
         """Test that Document.view_annotations() gets all the right annotations."""
         project = LocalTextProject()
-        document = project.test_documents[-1]
+        document = project.test_documents[-4]
```
da-konfuzio, line 2203: Probably easier to hardcode the document ID

docs/sdk/examples/file_splitting/file_splitting_without_model.md
```diff
@@ -0,0 +1,26 @@
...
+# initialize a class
+splitting_ai = SplittingAI(project_id=project.id_, category_id=test_document.category.id_)
+
+# gather Spans unique for the first Pages of the Documents in a chosen Category 
+unique_spans = splitting_ai.train()
+
+# propose a list of sub-Documents. If a test Document consists of a single file, return will consist of a list with the 
+# same Document.
+split_documents = splitting_ai.propose_split_documents(test_document, unique_spans)
+```
```
da-konfuzio: I think this is outdated right?

da-konfuzio:
Looks good to me, only left some minor comments. :slightly_smiling_face: 

Please solve conflicts and remove `docs/sdk/examples/file_splitting.md`, since the NN model is not included in this PR.

tests/test_data.py
iftwigs: the problem is that there are no IDs in LocalTextProject, if we assign them there will be an issue with trying to pull data from server and resulting in Errors (because there's no data for these Documents there)

docs/sdk/examples/file_splitting/file_splitting_without_model.md
iftwigs: yes it is

tests/trainer/test_file_splitting.py
iftwigs: I moved filtering by Category to the initialization level, so that is no longer needed.

da-konfuzio:
See this folder

https://github.com/konfuzio-ai/konfuzio-sdk/tree/master/tests/example_project_data/regex

The PR is trying to delete these files, please fix :slightly_smiling_face:

da-konfuzio (dismissed):

.gitignore
```diff
@@ -11,3 +11,5 @@ settings.py
 *.tar.gz
 *data_*
 .ipynb_checkpoints
+draft.ipynb
+testing_class/
```
da-konfuzio, line 15: Can be replaced with *.ipynb*

konfuzio_sdk/data.py
```diff
@@ -2054,6 +2061,7 @@ def download_document_details(self):
                 json.dump(data["pages"], f, indent=2, sort_keys=True)
         else:
             raise NotImplementedError
+            raise NotImplementedError
```
da-konfuzio, line 2064: duplicated line

setup.py
```diff
@@ -76,6 +76,9 @@
     install_requires=[
         'cloudpickle==2.0.0',  # Used to pickle objects
         'filetype==1.0.7',  # Used to check that files are in the correct format
+        'dill==0.3.2',  # Used to pickle objects
+        'keras==2.9.0',
+        'Keras-Preprocessing==1.1.2',
```
da-konfuzio, line 81: These wouldn't be used by the fallback logic so we can take them out for the moment

setup.py
```diff
@@ -85,7 +88,11 @@
         'requests',  # todo add ==2.24.0 ?
         'regex>=2020.6.8',  # re module but better
         'tabulate==0.8.7',  # Used to pretty print DataFrames
+        'tensorflow==2.9.1',
+        'torch==1.12.1',
+        'transformers==4.21.2',
         'tqdm',
+        'opencv-python==4.6.0.66',
```
da-konfuzio, line 95: These wouldn't be used by the fallback logic so we can take them out for the moment

tests/test_evaluate.py
```diff
@@ -945,7 +948,14 @@ def test_false_positive(self):
...
         project = LocalTextProject()
-        evaluation = Evaluation(documents=list(zip(project.documents, project.documents)))
+        documents_without_category_for_filesplitting = (
+            project.get_category_by_id(1).documents() + project.get_category_by_id(2).documents()
+        )
+        evaluation = Evaluation(
+            documents=list(
+                zip(documents_without_category_for_filesplitting, documents_without_category_for_filesplitting)
+            )
+        )
```
da-konfuzio, line 958: This is in the wrong section, probably you wanted to add it to TestEvaluationFileSplitting

docs/sdk/examples/evaluation/index.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+It is also possible to look at the metrics calculated by each Category independently. For this, pass `search=YOUR_CATEGORY_HERE` when calling the wanted metric's method: 
+```
+print(evaluation.fn(search=YOUR_CATEGORY_HERE))
+``` 
+
+### Example of evaluation input and output 
+
+Suppose in our test dataset we have 2 Documents of 2 Categories: one 3-paged, consisting of a single file (-> it has only one ground-truth first Page) of a first Category, and one 5-paged, consisting of three files: two 2-paged and one 1-paged (-> it has three ground-truth first Pages), of a second Category.
```
da-konfuzio, line 42: It would be good to show it visually with a table like the one you created here https://github.com/konfuzio-ai/konfuzio-sdk/blob/master/docs/sdk/examples/file-splitting-class/image.png

docs/sdk/examples/evaluation/index.md
```diff
@@ -0,0 +1,57 @@
...
+
+We need to pass a list of tuples of `(ground_truth_doc, predicted_doc)` into the `FileSplittingEvaluation` class, so we have to run both Documents' Pages through prediction before that and create a list for the input – `ground_truth_docs` would be original Documents and `predicted_docs` would be their copies with Pages predicted to be first or non-first.
+
+Suppose we ran the evaluation and received high results: only one first Page was predicted as non-first and the rest were predicted first/non-first correctly. Our results could be reflected in a following table:
+
```
da-konfuzio, line 47: Can you add a code example that would generate these scores?

tests/test_evaluate.py
```diff
@@ -1240,3 +1247,104 @@ def test_evaluation_calculator_none_values(self):
...
+        for category in self.file_splitting_model.categories:
+            cur_non_first_page_spans = []
+            for doc in category.documents():
+                for page in doc.pages():
+                    if page.number > 1:
+                        cur_non_first_page_spans.append({span.offset_string for span in page.spans()})
+            if not cur_non_first_page_spans:
+                cur_non_first_page_spans.append(set())
+            true_non_first_page_spans = set.intersection(*cur_non_first_page_spans)
+            non_first_page_spans[category.id_] = true_non_first_page_spans
```
da-konfuzio, line 1282: Probably this is some leftover code from a previous version, it can be removed :slightly_smiling_face:

.gitignore
```diff
@@ -11,3 +11,4 @@ settings.py
 *.tar.gz
 *data_*
 .ipynb_checkpoints
+*.ipynb
```
da-konfuzio, line 14: Would it work if the following two lines:
```
.ipynb_checkpoints
*.ipynb
```
were replaced with just:
```
*.ipynb*
```
?

docs/sdk/examples/evaluation/index.md
```diff
@@ -0,0 +1,194 @@
...
+To see a certain metric after the class has been initialized, you can call a metric's method. 
+```
+print(evaluation.fn())
+```
+
+It is also possible to look at the metrics calculated by each Category independently. For this, pass `search=YOUR_CATEGORY_HERE` when calling the wanted metric's method: 
+```
+print(evaluation.fn(search=YOUR_CATEGORY_HERE))
+```
```
da-konfuzio, line 38: "also see the Python API documentation https://dev.konfuzio.com/sdk/sourcecode.html#evaluation" for more details

docs/sdk/examples/evaluation/index.md
```diff
@@ -0,0 +1,194 @@
...
+
+Suppose in our test dataset we have 2 Documents of 2 Categories: one 3-paged, consisting of a single file (-> it has only one ground-truth first Page) of a first Category, and one 5-paged, consisting of three files: two 2-paged and one 1-paged (-> it has three ground-truth first Pages), of a second Category.
+
+![first_document_example](image_1.png)  
+
+_First document_
+
+![second_document_example](image_2.png)
+
+_Second document_
```
da-konfuzio, line 50: Are you sure these links are not broken?

docs/sdk/examples/evaluation/index.md
da-konfuzio: Sorry, I meant some documents from the OFFLINE_PROJECT (or even generated virtually on the fly like LocalTextProject), with example application of the FileSplittingEvaluation, and print the scores :sweat_smile:

docs/sdk/examples/evaluation/index.md
```diff
@@ -0,0 +1,57 @@
...
+We need to pass a list of tuples of `(ground_truth_doc, predicted_doc)` into the `FileSplittingEvaluation` class, so we have to run both Documents' Pages through prediction before that and create a list for the input – `ground_truth_docs` would be original Documents and `predicted_docs` would be their copies with Pages predicted to be first or non-first.
+
+Suppose we ran the evaluation and received high results: only one first Page was predicted as non-first and the rest were predicted first/non-first correctly. Our results could be reflected in a following table:
+
+| TPs | FPs | FNs | precision | recall | F1    |
+| ---- | ---- | ----- | ---- | ---- |-------|
+| 3 | 0 | 1 | 1 | 0.75 | 0.85  |
+
+If we want to see evaluation results by Category, the output could be reflected in a following table:
+
```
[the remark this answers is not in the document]
da-konfuzio: Sorry, I meant some documents from the OFFLINE_PROJECT (or even generated virtually on the fly like LocalTextProject), with example application of the FileSplittingEvaluation, and print the scores :sweat_smile:

konfuzio_sdk/trainer/file_splitting.py
```diff
@@ -0,0 +1,239 @@
...
+        """
+
+
+class ContextAwareFileSplittingModel(AbstractFileSplittingModel):
+    """Fallback definition of a File Splitting Model."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialize the ContextAwareFileSplittingModel."""
+        self.train_data = None
+        self.test_data = None
```
da-konfuzio, line 61: Rename
train_data -> documents
test_data -> test_documents
(to follow the same naming already present in Project, Category, and ExtractionAI)

konfuzio_sdk/trainer/file_splitting.py
```diff
@@ -0,0 +1,239 @@
...
+        :param page: A Page to receive first or non-first label.
+        :type page: Page
+        :return: A Page with or without is_first_page label.
+        """
+        for category in self.categories:
+            intersection = {span.offset_string for span in page.spans()}.intersection(
+                self.first_page_spans[category.id_]
+            )
+            if len(intersection) > 0:
+                page.is_first_page = True
```
da-konfuzio, line 136: We can `break` as soon as we find a match, right? :smile:

konfuzio_sdk/trainer/file_splitting.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def __init__(self, model=""):
+        """
+        Initialize the class.
+
+        :param model: A path to an existing .cloudpickle model or to a previously trained instance of
+        ContextAwareFileSplittingModel().
+        """
+        self.tokenizer = ConnectedTextTokenizer()
+        if model is str:
```
da-konfuzio, line 151: Please notice that if I run `"this is a string" is str` or even `"" is str` I get `False`

konfuzio_sdk/trainer/file_splitting.py
```diff
@@ -0,0 +1,239 @@
...
+        :type use_training_docs: bool
+        :return: Evaluation information for the filesplitting context-aware logic.
+        """
+        evaluation_list = []
+        if not use_training_docs:
+            evaluation_docs = self.model.test_data
+        else:
+            evaluation_docs = self.model.train_data
+        for doc in evaluation_docs:
+            doc.pages()[0].is_first_page = True
```
da-konfuzio, line 232: If the document is from Training or Test set, it should already have the attribute `is_first_page` set to True (you did this in data.py)

konfuzio_sdk/trainer/file_splitting.py
```diff
@@ -0,0 +1,239 @@
...
+            pred = self.tokenizer.tokenize(deepcopy(doc))
+            for page in pred.pages():
+                if self.model.predict(page).is_first_page:
+                    page.is_first_page = True
+            evaluation_list.append((doc, pred))
```
da-konfuzio, line 237: Tokenization and deepcopy should happen inside of
- `_suggest_first_pages(self, document: Document) -> Document`
- `_suggest_page_split(self, document: Document) -> List[Document]`
- (indeed it correctly happens in those methods)

Thus the goal should be that within evaluate_full, we do:
```python
for doc in evaluation_docs:
    predicted_doc = self.propose_split_documents(document=document, return_pages=True)
    evaluation_list.append((document, predicted_doc))
```

:hugs:

notebooks/.gitignore
```diff
@@ -0,0 +1,2 @@
+draft.ipynb
+*.pickle
```
da-konfuzio, line 2: Do you think we really need this additional .gitignore? Perhaps only the *pkl

konfuzio_sdk/trainer/file_splitting.py
da-konfuzio: Please update all the code examples in all the markdown files (not only docs/sdk/examples/evaluation/index.md), that the developer should never tokenize and deepcopy, as it happens in our internal methods

tests/test_evaluate.py
```diff
@@ -1240,3 +1247,93 @@ def test_evaluation_calculator_none_values(self):
...
+    @classmethod
+    def setUpClass(cls) -> None:
+        """Initialize the tested class."""
+        cls.project = LocalTextProject()
+        cls.file_splitting_model = ContextAwareFileSplittingModel()
+        cls.file_splitting_model.categories = cls.project.categories
+        cls.file_splitting_model.train_data = [
+            document for category in cls.file_splitting_model.categories for document in category.documents()
+        ]
+        cls.file_splitting_model.test_data = cls.project.get_category_by_id(3).test_documents()
```
da-konfuzio, line 1264: Just nitpicking but why did you not use the same list comprehension in the test documents?

tests/test_evaluate.py
```diff
@@ -1240,3 +1247,93 @@ def test_evaluation_calculator_none_values(self):
...
+        pred = splitting_ai.propose_split_documents(self.test_document, return_pages=True)
+        documents = [[ground_truth, pred]]
+        evaluation = FileSplittingEvaluation(documents)
+        assert evaluation.tp() == 3
+        assert evaluation.fp() == 0
+        assert evaluation.fn() == 0
+        assert evaluation.tn() == 2
+        assert evaluation.precision() == 1.0
+        assert evaluation.recall() == 1.0
+        assert evaluation.f1() == 1.0
```
da-konfuzio, line 1286: Very good test!

Can you add one where the resulting metrics are not 100%?

tests/trainer/.gitignore
```diff
@@ -0,0 +1 @@
+*.pickle
```
da-konfuzio, line 1: Are you sure it should be .pickle and not .pkl? Or .cloudpickle? Please look at the file extension used everywhere else in the SDK and unify it

tests/trainer/test_file_splitting.py
```diff
@@ -0,0 +1,116 @@
...
+    def test_suggest_first_pages(self):
+        """Test SplittingAI's suggesting first Pages."""
+        splitting_ai = SplittingAI(self.file_splitting_model)
+        pred = splitting_ai.propose_split_documents(self.test_document, return_pages=True)
+        for page in pred.pages():
+            if page.number in (1, 3, 5):
+                assert page.is_first_page
+            else:
+                assert not page.is_first_page
+        for item in self.project.model_folder:
```
da-konfuzio, line 89: This for loop is iterating over characters in the string of the folder path

It's not iterating over the files in the folder :hugs:

docs/sdk/examples/evaluation/index.md
da-konfuzio: For example I see here https://dev.konfuzio.com/sdk/examples/examples.html#splitting-for-multi-file-documents-step-by-step-guide

The "Multi-file Document Example" is not displaying

tests/test_evaluate.py
iftwigs: actually it seems even better to mimick the logic from test documents in the train documents, because it does not make sense iterating over Documents from Categories 1&2 (they don't even have pagination, so no use for training).

iftwigs: actually, it makes more sense to mimick the logic from test_documents in the train documents, as there isn't even pagination in Documents from Categories 1&2.

docs/sdk/examples/evaluation/index.md
```diff
@@ -0,0 +1,176 @@
...
+print(evaluation.fp(search=CATEGORY_1), evaluation.fp(search=CATEGORY_2)) # returns: 0 0
+print(evaluation.fn(search=CATEGORY_1), evaluation.fn(search=CATEGORY_2)) # returns: 0 1
+print(evaluation.precision(search=CATEGORY_1), evaluation.precision(search=CATEGORY_2)) # returns: 1 1
+print(evaluation.recall(search=CATEGORY_1), evaluation.recall(search=CATEGORY_2)) # returns: 1 0.66
+print(evaluation.f1(search=CATEGORY_1), evaluation.f1(search=CATEGORY_2)) # returns: 1 0.79
+```
+
+the output could be reflected in a following table:
+
+| Category | TPs | TNs | FPs | FNs | precision | recall | F1   |
```
atraining, line 173: do you want to add a helpfer function that create this table?

might be helpful for all Evaluation classes

docs/sdk/examples/file_splitting.md
```diff
@@ -0,0 +1,64 @@
+## Split Multi-File Documents into Separate Files
```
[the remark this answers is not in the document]
iftwigs: this file is now deleted from the branch, is it possible that you've viewed an outdated version of the repo?

konfuzio_sdk/trainer/file_splitting.py
```diff
@@ -0,0 +1,158 @@
...
+
+    :param documents: A group of Documents in Pages of which to search for the common Spans.
+    :type documents: list
+    :param category: A Category for which the search is done.
+    :type category: Category
+    :param tokenizer: A tokenizer to split Documents into Spans.
+    :return: A set of unique Spans.
+    """
+    # we can define possible modes of usage and suggest using them for suitable occurences
+    if kwargs['mode'] == 'file_splitting':
```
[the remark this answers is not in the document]
iftwigs: this method is deleted in a current version of the branch

konfuzio_sdk/trainer/file_splitting.py
```diff
@@ -0,0 +1,158 @@
...
+            first_page_spans.append(set())
+        true_first_page_spans = set.intersection(*first_page_spans)
+        if not not_first_page_spans:
+            not_first_page_spans.append(set())
+        true_not_first_page_spans = set.intersection(*not_first_page_spans)
+        true_first_page_spans = true_first_page_spans - true_not_first_page_spans
+    return true_first_page_spans
+
+
+class BaseCommonFeatureSearcher(ABC):
```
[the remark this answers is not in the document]
iftwigs: this class is deleted in the current version of the repo

docs/sdk/examples/file-splitting-class/out-of-the-box-usage.md
```diff
@@ -4,7 +4,6 @@ Let's see how to use the `konfuzio_sdk` to automatically split documents consist
 several files. We will be using a pre-built class SplittingAI. The class implements a context-aware rule-based logic 
 that requires no training.
 
-
 ```python
 from konfuzio_sdk.data import Project
```
atraining, line 8: ... please use descriptive file names if or keep it in index.md

add tests for source code in md files!

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -141,43 +143,57 @@ def fit(self, *args, **kwargs) -> dict:
         return first_page_spans
 ```
 
-Secondly, we define `save()` method. For saving, we use cloudpickle utility and compression with bz2. We save a set of 
-unique first-page Spans as the model.
+Secondly, we define `save()` method. For saving, it is possible to choose one of the two options: to save 
+`first_page_spans` as JSON (`save_json` has to be set to True) or to save a whole instance of the class as a pickle, 
+using parent class's `save()` method (`save_json` has to be set to False). A flag `include_konfuzio` enables pickle serialization as a value, not as a reference (for more info, read 
+[this](https://github.com/cloudpipe/cloudpickle#overriding-pickles-serialization-mechanism-for-importable-constructs)).
+`max_ram` is used to define the maximum memory usage condition to save model; `reduce_weight` allows for removing
```
atraining, line 150: wrong context, max ram should be discussed on a higher level in the documentation and moved from here

it applies to any pickle save and load strategy where infrastructure is differen in training or inference Virtual Machines

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -141,43 +143,57 @@ def fit(self, *args, **kwargs) -> dict:
...
-    pkl_file_path = os.path.join(model_path, f'{get_timestamp()}_first_page_spans.pkl')
-    with open(temp_pkl_file_path, 'wb') as f:
-        cloudpickle.dump(self.first_page_spans, f)
-    with open(temp_pkl_file_path, 'rb') as input_f:
-        with bz2.open(pkl_file_path, 'wb') as output_f:
-            shutil.copyfileobj(input_f, output_f)
-    os.remove(temp_pkl_file_path)
-    return pkl_file_path
+    def load_json(self, model_path=""):
+        with open(model_path, 'r') as f:
```
atraining, line 173: see data.py how to ensure correct orgin of path

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -141,43 +143,57 @@ def fit(self, *args, **kwargs) -> dict:
...
 Lastly, we define `predict()` method that uses resulting `first_page_spans` gathered from `fit()`. A Page is accepted as
  an input and its Span set is checked for containing first-page Spans for each of the Categories. If there has been at 
 least one intersection, a Page is predicted to be first; if there's no such intersections, it's predicted non-first.
 ```python
-def predict(self, page: Page) -> Page:
-        for category in self.categories:
+ def predict(self, page: Page) -> Page:
+     for category in self.categories:
             intersection = {span.offset_string for span in page.spans()}.intersection(
                 self.first_page_spans[category.id_]
```
atraining, line 187: span.offset_string for span in page.spans() -> might be easier testable as Page mth

self.first_page_spans[category.id_] -> seems to contain no real no Spans but is named like this

code section looks strange when: dictionary to keep first page spans, and not working with spans but with strings

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -280,35 +308,46 @@ class ContextAwareFileSplittingModel(AbstractFileSplittingModel):
...
-        pkl_file_path = os.path.join(model_path, f'{get_timestamp()}_first_page_spans.pkl')
-        with open(temp_pkl_file_path, 'wb') as f:
-            cloudpickle.dump(self.first_page_spans, f)
-        with open(temp_pkl_file_path, 'rb') as input_f:
-            with bz2.open(pkl_file_path, 'wb') as output_f:
-                shutil.copyfileobj(input_f, output_f)
-        os.remove(temp_pkl_file_path)
-        return pkl_file_path
+        if save_json:
+            path = self.output_dir + f'/{get_timestamp()}_first_page_spans.json'
```
atraining, line 326: not unique for different categories?

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -280,35 +308,46 @@ class ContextAwareFileSplittingModel(AbstractFileSplittingModel):
...
+        Load JSON with previously gathered first_page_spans.
+
+        :param model_path: Path for the JSON.
+        :type model_path: str
+        """
+        with open(model_path, 'r') as f:
+            spans = json.load(f)
+        # converting str category.id_ values to back int because JSON converts them to str
+        spans = {int(k): v for k, v in spans.items()}
+        self.first_page_spans = spans
```
atraining, line 345: why not working on real Spans.

konfuzio_sdk/trainer/file_splitting.py
```diff
@@ -0,0 +1,296 @@
...
+        for page_i, split_i in enumerate(suggested_splits):
+            if page_i == 0:
+                split_docs.append(self._create_doc_from_page_interval(document, first_page, split_i))
+            elif page_i == len(split_docs):
+                split_docs.append(self._create_doc_from_page_interval(document, split_i, last_page))
+            else:
+                split_docs.append(self._create_doc_from_page_interval(document, suggested_splits[page_i - 1], split_i))
+        return split_docs
+
+    def propose_split_documents(
```
atraining, line 256: review function, as it returns a instance or a list... hard to handle and test

konfuzio_sdk/evaluate.py
```diff
@@ -437,3 +453,202 @@ def get_wrong_vertical_merge(self):
...
+    """Evaluate the quality of the filesplitting logic."""
+
+    def __init__(self, documents: List[Tuple[Document, Document]]):
+        """
+        Initialize and run the metrics calculation.
+
+        :param documents: A list of Document pairs – first one is ground truth, second is the prediction.
+        :type documents: list
+        """
+        self.documents = documents
```
atraining, line 468: please test this init, and make sure the init does basic logic checks

konfuzio_sdk/evaluate.py
```diff
@@ -437,3 +453,202 @@ def get_wrong_vertical_merge(self):
...
+            raise ZeroDivisionError("TP and FP are zero.")
+        if tp + fn != 0:
+            recall = tp / (tp + fn)
+        else:
+            raise ZeroDivisionError("TP and FN are zero.")
+        if precision + recall != 0:
+            f1 = 2 * precision * recall / (precision + recall)
+        else:
+            raise ZeroDivisionError("FP and FN are zero.")
+        self.project = self.documents[0][0].project
```
atraining, line 500: issues will appear when documents in the first list belong to a different project than documents in the second list

you need to restrict the data which can enter and you can manage both lists of document as separate variable if needed

it's hard to get that self.documents is not a list of docs but a list of tuples of docs, and its unclear if its predicted or proposed / ground truth document first.

konfuzio_sdk/evaluate.py
```diff
@@ -437,3 +453,202 @@ def get_wrong_vertical_merge(self):
...
+            'precision': {},
+            'recall': {},
+            'f1': {},
+        }
+        for category in categories:
+            tp = 0
+            fp = 0
+            fn = 0
+            tn = 0
+            for ground_truth, prediction in [
```
atraining, line 528: help yourself and unwarp this in the init

konfuzio_sdk/evaluate.py
```diff
@@ -437,3 +453,202 @@ def get_wrong_vertical_merge(self):
...
+                raise ZeroDivisionError("TP and FP are zero.")
+            if tp + fn != 0:
+                recall = tp / (tp + fn)
+            else:
+                raise ZeroDivisionError("TP and FN are zero.")
+            if precision + recall != 0:
+                f1 = 2 * precision * recall / (precision + recall)
+            else:
+                raise ZeroDivisionError("FP and FN are zero.")
+            self.evaluation_results_by_category['tp'][category.id_] = tp
```
atraining, line 554: looks odd, was not done by the class on top which used the same process to evaluate extraction quality

konfuzio_sdk/evaluate.py
```diff
@@ -437,3 +453,202 @@ def get_wrong_vertical_merge(self):
...
+
+    def tn(self, search: Category = None) -> int:
+        """
+        Return non-first Pages predicted as non-first.
+
+        :param search: display true negatives within a certain Category.
+        :type search: Category
+        :raises KeyError: When the Category in search is not present in the Project from which the Documents are.
+        """
+        return self._query('tn', search)
```
atraining, line 624: 'tn' working is not in line with wording in the class above, it called it true_negative IMHO

docs/sdk/examples/file-splitting-class/out-of-the-box-usage.md
iftwigs: changed the name to a descritpive one. 
about the tests for the code in md files – Davide pointed to a [ticket](https://git.konfuzio.com/konfuzio/objectives/-/issues/10017) for that, covering this with tests is in its scope too.

docs/sdk/examples/file-splitting-class/index.md
iftwigs: no. it can be done for multiple Categories simultaneously and is saved as a dictionary with Categories' IDs as keys and lists of unique Spans' strings as values.

docs/sdk/examples/file-splitting-class/index.md
iftwigs: when we compare Spans instead of comparing their strings, we are unlikely to find intersections between sets of Spans across different Pages, because even two Spans with the same string can have different start_offset and end_offset, for instance, and thus would not be considered "equal" -> the resulting set of unique first-Page Spans would be empty. after that was noted, we switched to using strings instead of Spans themselves.

docs/sdk/examples/file-splitting-class/index.md
iftwigs: we got rid of JSON saving, so this part is out now too.

docs/sdk/examples/file-splitting-class/index.md
iftwigs: renamed all first_page_spans into first_page_strings; about making it a Page method – decided on review with Davide and Flo to leave it as it is.

tests/trainer/test_file_splitting.py
```diff
@@ -0,0 +1,219 @@
+"""Test SplittingAI and the model's training, saving and prediction."""
+import os
```
atraining, line 2: there are many binaries pkl checked in --> @zypriafl

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -2,80 +2,77 @@
...
-        self.categories = None
-        self.first_page_spans = None
-        self.tokenizer = None
-        
+    def __init__(self, categories: List[Category], tokenizer, *args, **kwargs):
+        super().__init__(categories=categories)
+        self.name = self.__class__.__name__
+        self.output_dir = self.project.model_folder
+        self.tokenizer = tokenizer
+        self.path = None
```
da-konfuzio, line 62: Define self.path in the Abstract class so the developer doesn't have to do this explicitly in the derived class

da-konfuzio: same with self.output_dir

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -101,113 +98,64 @@ test_document.spans[0].offset_string
...
-        return first_page_spans
+            cur_first_page_strings = category.exclusive_first_page_strings(tokenizer=self.tokenizer)
+            if not cur_first_page_strings:
+                if allow_empty_categories:
+                    logger.warning(
+                        f'No exclusive first-page strings were found for {category}, so it will not be used '
+                        f'at prediction.'
+                    )
+                else:
+                    raise ValueError(f'No exclusive first-page strings were found for {category}.')
```
da-konfuzio, line 116: Can you add an explanation that there are private attributes in the Category which are getting set by this method, such that the information is not lost when this local variable `cur_first_page_strings` is removed?

When reading this, it may look like the variable containing the result is lost.

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -101,113 +98,64 @@ test_document.spans[0].offset_string
...
- an input and its Span set is checked for containing first-page Spans for each of the Categories. If there has been at 
-least one intersection, a Page is predicted to be first; if there's no such intersections, it's predicted non-first.
+Lastly, we define `predict()` method. The method accepts a Page as an input and checks its Span set for containing 
+first-page strings for each of the Categories. If there is at least one intersection, the Page is predicted to be a 
+first Page. If there are no intersections, the Page is predicted to be a non-first Page.
 ```python
-def predict(self, page: Page) -> Page:
+    def predict(self, page: Page) -> Page:
+        for category in self.categories:
+            if not category.exclusive_first_page_strings:
```
da-konfuzio, line 125: It's a method, not a property. Can you add a comment explaining that the result of this computation is cached within Category, such that it's computationally cheap to check the value of the internal variable by calling this method?

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -222,137 +170,90 @@ splitting_ai = SplittingAI(model)
 # SplittingAI can be ran in two modes: returning a list of sub-Documents as the result of the input Document
 # splitting or returning a copy of the input Document with Pages predicted as first having an attribute
 # "is_first_page". The flag "return_pages" has to be True for the latter; let's use it
-new_document = splitting_ai.predict(test_document, return_pages=True)
+new_document = splitting_ai.propose_split_documents(test_document, return_pages=True)
```
da-konfuzio, line 173: Here we can explain that SpittingAI is a higher level interface compared to AbstractFileSplittingModel, thus it takes as an input a Document rather than individual Pages, because it internally uses the per Page prediction method that was customized. Please add the print statements as above so that the result can be shown.

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -222,137 +170,90 @@ splitting_ai = SplittingAI(model)
...
 ```
+Note that any custom FileSplittingAI (derived from `AbstractFileSplittingModel` class) requires having the following 
+methods implemented:
+- `__init__` to initialize key variables required by the custom AI;
+- `fit` to define architecture and training that the model undergoes, i.e. a certain NN architecture or a custom 
+- hardcoded logic
+- `predict` to define how the model classifies Pages as first or non-first. **NB:** the classification needs to be ran on 
+the Page level, not the Document level – the result of classification is reflected in `is_first_page` attribute value, which
+is unique to the Page class and is not present in Document class. Pages with `is_first_page = True` become splitting 
+points, thus, each new sub-Document has a Page predicted as first as its starting point.
```
da-konfuzio, line 183: This can go at the beginning before we actually customize each method :slightly_smiling_face:

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1073,13 +1068,115 @@ def add_extractions_as_annotations(
...
+        :param keep_documents: To allow restoring Documents after reducing model's weight in case evaluation is needed.
+        :type keep_documents: bool
+        :param max_ram: Specify maximum memory usage condition to save model.
+        :raises MemoryError: When the size of the model in memory is greater than the maximum value.
+        :return: Path of the saved model file.
+        """
+        logger.info('Saving model')
+        if self.output_dir is None:
+            raise OSError("Specify output_dir before saving the model.")
+        pathlib.Path(self.output_dir).mkdir(parents=True, exist_ok=True)
```
atraining, line 1126: why not validated in the init?

tests/trainer/test_information_extraction.py
```diff
@@ -1194,27 +1200,35 @@ def test_load_model_corrupt_file():
...
-    with pytest.raises(TypeError, match="needs to be a Konfuzio Trainer instance"):
+    with pytest.raises(TypeError, match="not inheriting from the BaseModel class."):
         load_model(path)
 
 
 def test_load_ai_model():
     """Test loading of trained model."""
     project = Project(id_=None, project_folder=OFFLINE_PROJECT)
-    path = "trainer/2022-09-27-18-45-41_lohnabrechnung.pkl"
+    path = "trainer/2023-01-17-20-21-13_lohnabrechnung.pkl"
```
atraining, line 1210: why are there so many pkl checked in?

konfuzio_sdk/data.py
```diff
@@ -108,6 +109,13 @@ def __init__(
         self.width = self._original_size[0]
         self.height = self._original_size[1]
         self.image_path = os.path.join(self.document.document_folder, f'page_{self.number}.png')
+        if self.document.dataset_status in (2, 3):
+            if self.number == 1:
+                self.is_first_page = True
```
atraining, line 114: why is this only the case for data set status 2 and 3 ?

why is it not calulated as a property

konfuzio_sdk/data.py
```diff
@@ -108,6 +109,13 @@ def __init__(
...
+            else:
+                self.is_first_page = False
+        else:
+            self.is_first_page = None
```
atraining, line 118: why is the default not setup first and then overwritten, why are those lines not tested?

konfuzio_sdk/data.py
```diff
@@ -549,6 +559,50 @@ def add_label_set(self, label_set):
...
+            doc = deepcopy(doc)
+            doc = tokenizer.tokenize(doc)
+            for page in doc.pages():
+                if page.number == 1:
+                    cur_first_page_strings.append({span.offset_string for span in page.spans()})
+                else:
+                    cur_non_first_page_strings.append({span.offset_string for span in page.spans()})
+            doc.delete()
+        if not cur_first_page_strings:
+            cur_first_page_strings.append(set())
```
atraining, line 580: adding a empty set looks suspicious, can this be optimized

konfuzio_sdk/data.py
```diff
@@ -549,6 +559,50 @@ def add_label_set(self, label_set):
...
+        self._exclusive_first_page_strings = true_first_page_strings
+
+    def exclusive_first_page_strings(self, tokenizer) -> set:
+        """
+        Return a set of strings exclusive for first Pages of Documents within the Category.
+
+        :param tokenizer: A tokenizer to process Documents before gathering strings.
+        """
+        if self._exclusive_span_tokenizer is not None:
+            if tokenizer != self._exclusive_span_tokenizer:
```
atraining, line 595: why do we need to parse the tokenizer again if we would want to probably reject it anyhow, if it's not equal to the tokenizer before?

konfuzio_sdk/trainer/information_extraction.py
iftwigs: because we don't pass output_dir at the init, but we have the checks for this in child classes. and in case there's a new child class from the BaseModel that does not have the checks, we want to catch it here.

tests/trainer/test_information_extraction.py
iftwigs: because there was a re-generation of them, older ones were generated before BaseModel class was introduced and thus didn't pass inheritance checks.

konfuzio_sdk/data.py
iftwigs: because we assume that only these Documents are ground-truth in the sense of file splitting

iftwigs: about it not being a property: because there is a context where it is explicitly redefined

docs/sdk/examples/architecture_overview/index.md
```diff
@@ -0,0 +1,17 @@
+## Architecture overview
+
+we'll take a closer look at the most important ones and give you some examples of how they can be implemented 
+consequentially or individually in case you want to experiment.
```
da-konfuzio, line 4: It seems that the text was cut by mistake

konfuzio_sdk/data.py
iftwigs: if it is not equal to the tokenizer before, we just re-tokenize using a new one, we don't want to reject it.

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -2,80 +2,85 @@
...
-an assumption that we can go through all Pages of the Document and define splitting points. By splitting points, we mean 
-Pages that are similar in contents to the first Pages in the Documents. 
-Note: this approach only works with the Documents of the same Category and in the same language. 
+In this section, we will explore an easy method for identifying and separating Documents that may be included in a 
+single file. Our approach involves analyzing the contents of each Page and identifying similarities to the first Pages 
+of the Document. This will allow us to define splitting points and divide the Document into multiple sub-documents. It's
+important to note that this approach is only effective for Documents written in the same language and that the process 
+must be repeated for each Category.
 
 If you are unfamiliar with the SDK's main concepts (like Page or Span), you can get to know them on the Quickstart page.
```
da-konfuzio, line 19: link to quickstart

tests/test_data.py
```diff
@@ -2267,15 +2338,15 @@ def test_make_sure_pages_are_not_downloaded_automatically(self):
     def test_view_annotations(self):
         """Test that Document.view_annotations() gets all the right annotations."""
         project = LocalTextProject()
-        document = project.test_documents[-1]
+        document = project.test_documents[-7]
```
da-konfuzio, line 2341: Would it be better to encode the document id?

iftwigs: this is done purely because there is no document id, because otherwise there is a check for the document being online (and since it's not there, it crashes)

iftwigs: this is the case with all the Documents in LocalTextProject

tests/trainer/test_file_splitting.py
```diff
@@ -0,0 +1,220 @@
...
+                cur_first_page_strings = category.exclusive_first_page_strings(
+                    tokenizer=self.file_splitting_model.tokenizer
+                )
+                intersection = {span.offset_string.strip('\f').strip('\n') for span in page.spans()}.intersection(
+                    cur_first_page_strings
+                )
+                if len(intersection) > 0:
+                    page.is_first_page = True
+                    break
+            print(page.is_first_page, page.text)
```
da-konfuzio, line 110: Remove print :slightly_smiling_face:

tests/test_data.py
da-konfuzio: Arg, this will be hard to maintain.

But I realize what we can do in samples.py:
- in the constructor of LocalTextProject, immediately call `self.set_offline()`
- then add IDs to all the documents

da-konfuzio (dismissed):
LGTM pending resolution of remaining comments, it can be merged.

Then just add the usual description to the PR (mostly the documentation links, only one is linked at the moment but we created many markdown files) :wink:

da-konfuzio (approved):

PR #156, opened 2022-11-14
Changed: Log File creation can be disabled via an environment variable

da-konfuzio:
An environment variable (or an entry to the `.env` file) can now be used for disabling the creation of the log file.

### DIFF
--- a/konfuzio_sdk/settings_importer.py
+++ b/konfuzio_sdk/settings_importer.py
@@ -27,7 +27,7 @@ LOG_FORMAT = (
 
 handlers = [logging.StreamHandler()]
 
-if config('LOG_TO_FILE', default=True):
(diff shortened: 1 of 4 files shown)


konfuzio_sdk/settings_importer.py
```diff
@@ -27,7 +27,7 @@
 
 handlers = [logging.StreamHandler()]
 
-if config('LOG_TO_FILE', default=True):
+if config('LOG_TO_FILE', cast=bool, default=True):
```
atraining, line 30: should we stick to the standards how to configure logging https://docs.python-guide.org/writing/logging/#example-configuration-via-an-ini-file

da-konfuzio: Thanks, tracking on internal ticket #10092

zypriafl (approved):

PR #164, opened 2022-11-19
This draft allows to init Annotations which do not have any Span

zypriafl:


konfuzio_sdk/data.py
```diff
@@ -1291,7 +1291,9 @@ def __init__(
                 if "start_offset" in bbox.keys() and "end_offset" in bbox.keys():
                     Span(start_offset=bbox["start_offset"], end_offset=bbox["end_offset"], annotation=self)
                 else:
-                    ValueError(f'SDK cannot read bbox of Annotation {self.id_} in {self.document}: {bbox}')
+                    # Add Span that does not include any character.
+                    span = Span(annotation=self, start_offset=0, end_offset=0)
```
atraining, line 1295: I would not do it, this can have an impact on our training data without our knowledge,

atraining: we are strict at any point until know, this brings back vague appraoches

PR #176, opened 2022-12-06
Added: Monitoring of RAM usage

samuel500:
This adds a number of tests and improvements to monitor and reduce memory usage of Konfuzio RFExtractionAI models.


konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -2395,8 +2400,12 @@ def fit(self) -> RandomForestClassifier:
 
         self.clf.fit(self.df_train[self.label_feature_list], self.df_train['target'])
 
+        logger.info(f"Size of Label classifier: {asizeof.asizeof(self.clf)/1000} KB.")
+
         self.fit_label_set_clf()
 
+        logger.info(f"Size of LabelSet classifier: {asizeof.asizeof(self.label_set_clf)/1000} KB.")
```
da-konfuzio, line 2407: I see testcases about sizes of various things except the `self.label_set_clf`

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -2382,6 +2385,8 @@ def feature_function(
         else:
             raise NotImplementedError  # = pandas.DataFrame()
 
+        logger.info(f"Size of feature dict {asizeof.asizeof(df_real_list)/1000} KB.")
```
da-konfuzio, line 2388: For the purpose of having an easy and "Konfuzio SDK official" way of measuring sizes of things (or just this RFExtractionAI stuff for the moment), we can have some helper function such as `utils.py::memory_size_of(obj) -> int` (in bytes), which wraps `asizeof.asizeof` or other suitable function.

Ideally we should be able to get `memory_size_of` a RFExtractionAI instance and `memory_size_of(df_real_list)` and so on.

PR #184, opened 2022-12-09
Added: Options to reduce RAM usage and file size of Extraction AIs

samuel500:
This reduces the weight of trained Extraction AI models on disk and in memory by removing various data structures not needed for inference.

The `Trainer.save` method now allows to control saving behaviour with the `reduce_weight` and `keep_documents` options.

For more details see the documentation for [RFExtractionAI](https://dev.konfuzio.com/sdk/sourcecode.html#extraction-ai) and the various [Konfuzio data structures](https://dev.konfuzio.com/sdk/sourcecode.html#data).


konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1679,13 +1555,13 @@ def save(self, output_dir: str = None, include_konfuzio=True, reduce_weight=Fals
 
         logger.info('Deleting cloudpickle file')
         # then delete cloudpickle file
-        os.remove(temp_pkl_file_path)
+        # os.remove(temp_pkl_file_path)
```
da-konfuzio, line 1558: Does this make sense?

samuel500: Good catch. I had commented it out while running some tests and forgot to put it back.

da-konfuzio (approved):

PR #188, opened 2022-12-14
Fixed: A small bug that blocked the extraction on blank pages and blank documents

samuel500:
This will make it so that no error is raised when the tokenizer cannot create any Annotations, and it only returns an empty document.


konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -2158,7 +2158,7 @@ def extract(self, document: Document) -> Document:
         self.tokenizer.tokenize(inference_document)
         if not inference_document.spans():
             logger.error(f'{self.tokenizer} does not provide Spans for {document}')
-            raise NotImplementedError('No error handling when Spans are missing.')
+            return inference_document
```
da-konfuzio, line 2161: Why not add a testcase? :slightly_smiling_face:

tests/trainer/test_information_extraction.py
```diff
@@ -479,6 +479,22 @@ def test_extraction_without_clf(self):
...
+        pipeline.clf = RandomForestClassifier(max_depth=5, n_estimators=10, max_features=1)
+        X, y = make_classification(
+            n_samples=1000, n_features=4, n_informative=2, n_redundant=0, random_state=0, shuffle=False
+        )
+        pipeline.clf.fit(X, y)
+
+        virt_doc = pipeline.extract(document)
+
+        assert virt_doc.annotations(use_correct=False) == []
+
```
da-konfuzio, line 497: Nice! But what about if there really is no text or pages in the document?

da-konfuzio (approved):

PR #189, opened 2022-12-16
Added: Save method for saving pickle files of File Splitting AI and Extraction AI refactored

iftwigs:
This change provides a parent BaseModel class with a universal save method defined for all child classes, and allows for saving ContextAwareFileSplittingModel as a pickled class instance or as a JSON of first-Page Spans that can be used for proper multi-file Document splitting.

Find the following documentation:

- [Code Examples: FileSplittingEvaluation class](https://dev.konfuzio.com/sdk/examples/examples.html#filesplittingevaluation-class)


konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1324,7 +1325,97 @@ def add_extractions_as_annotations(
...
+        :type include_konfuzio: bool
+        :param model_type: file_splitting or extraction_ai.
+        :type model_type: str
+        :param reduce_weight: Remove all non-strictly necessary parameters before saving.
+        :param max_ram: Specify maximum memory usage condition to save model.
+        :raises MemoryError: When the size of the model in memory is greater than the maximum value.
+        :return: Path of the saved model file.
+        """
+        logger.info('Saving model')
+        logger.info(f'{include_konfuzio=}')
```
atraining, line 1348: should this not log the version too ? this variable is missing then

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1324,7 +1325,97 @@ def add_extractions_as_annotations(
...
+        if include_konfuzio:
+            cloudpickle.register_pickle_by_value(konfuzio_sdk)
+            # todo register all dependencies?
+        pathlib.Path(output_dir).mkdir(parents=True, exist_ok=True)
```
atraining, line 1352: to follow the practice of the other classes this would go in init

why is output dir the working dir and not a init variable, at least to be consistent

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1324,7 +1325,97 @@ def add_extractions_as_annotations(
...
+        if model_type == "file_splitting":
+            temp_pkl_file_path = os.path.join(output_dir, f'{get_timestamp()}_first_page_spans_tmp.cloudpickle')
+            pkl_file_path = os.path.join(output_dir, f'{get_timestamp()}_first_page_spans.pkl')
```
atraining, line 1355: why do we have hardcoded different file name endings if the pickling is the same?

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1324,7 +1325,97 @@ def add_extractions_as_annotations(
...
+        elif model_type == "extraction_ai":
```
atraining, line 1356: can we have sth better than a string model type, rather the inheritated class name?

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1324,7 +1325,97 @@ def add_extractions_as_annotations(
...
+
+            if max_ram and asizeof.asizeof(self) > max_ram:
+                raise MemoryError(f"AI model memory use ({asizeof.asizeof(self)}) exceeds maximum ({max_ram=}).")
+
+            sys.setrecursionlimit(99999999)  # ?
+
+            logger.info('Getting save paths')
+
+            if not output_dir:
+                output_dir = self.category.project.model_folder
```
atraining, line 1388: a lot of custom paths until we end here,  why is this not the default and it can be overwritten when calling the fct

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1324,7 +1325,97 @@ def add_extractions_as_annotations(
...
+
+            temp_pkl_file_path = os.path.join(output_dir, f'{get_timestamp()}_{self.category.name.lower()}.cloudpickle')
```
atraining, line 1390: self.category can be None

tests/trainer/test_information_extraction.py
```diff
@@ -225,14 +225,18 @@ def test_03_fit(self) -> None:
...
         with pytest.raises(MemoryError):
-            self.pipeline.pipeline_path = self.pipeline.save(include_konfuzio=False, max_ram="5MB")
+            self.pipeline.pipeline_path = self.pipeline.save(
+                model_type="extraction_ai", include_konfuzio=False, max_ram="5MB"
+            )
 
         self.project._max_ram = "5MB"
         with pytest.raises(MemoryError):
-            self.pipeline.pipeline_path = self.pipeline.save(include_konfuzio=False)
+            self.pipeline.pipeline_path = self.pipeline.save(model_type="extraction_ai", include_konfuzio=False)
```
atraining, line 234: see comments above, it feels strange to parse the self.pipeline a model type which should be defined by the instance self.pipeline

konfuzio_sdk/trainer/information_extraction.py
iftwigs: this cannot be made a default, because category is not a common attribute among the child classes of BaseModel (in ContextAwareFileSplittingModel it is categories, since we can pass multiple into it). i set up two default options depending on the child class.

konfuzio_sdk/trainer/information_extraction.py
iftwigs: according to Davide, it is not an intended usage to run the ExtractionAI if self.category is None.

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -42,10 +42,7 @@ find in the full code block in the lower part of this page. A class itself is al
```
da-konfuzio, line 42: Small typo (not here but at the beginning of the MD file, I can't comment there because that section was not modified by this PR): "we can have more than one actual Document in a stream pf pages"

"pf" instead of "of" :wink:

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -60,22 +57,24 @@ from konfuzio_sdk.utils import get_timestamp
...
-The class inherits from `AbstractFileSplittingModel`. Attributes are set to `None` to be assigned explicitly upon usage 
-later. `train_data` and `test_data` will be used for "training" (gathering unique first-page Spans from the Documents) 
-and testing respectively; `categories` will define groups of Documents within which the "training" will take place;
-`first_page_spans` will be defined with the result of `fit()` method running; `tokenizer` will be used for going through 
-the Document's text and separating it into Spans. We will use it to process training and testing Documents, as well as 
-any Document that will undergo splitting. This is done to ensure that texts in all of the Documents are split using the 
-same logic (particularly tokenization by separating on `\n` whitespaces by ConnectedTextTokenizer, which is used in the 
-example in the end of the page) and it will be possible to find common Spans. 
+The class inherits from `AbstractFileSplittingModel`, so we run `super().__init__()` for proper inheritance of the 
+attributes. Attributes are set to `None` to be assigned explicitly upon usage later. `train_data` and `test_data` will be
```
da-konfuzio, line 70: still talks about train_data and test_data but we renamed them to documents and test_documents (do you need to merge 9824-file-splitting-fallback-logic into this branch?)

docs/sdk/examples/file-splitting-class/index.md
```diff
@@ -141,39 +140,45 @@ def fit(self, *args, **kwargs) -> dict:
...
+ def predict(self, page: Page) -> Page:
+     for category in self.categories:
             intersection = {span.offset_string for span in page.spans()}.intersection(
                 self.first_page_spans[category.id_]
             )
             if len(intersection) > 0:
                 page.is_first_page = True
+                break
         return page
 ```
```
da-konfuzio, line 184: When we say "A quick example of the class's usage" let's add that any custom File Splitting AI (derived from AbstractFileSplittingModel like we are doing with ContextAwareFileSplittingModel as an example) will require these methods to be defined and why they are needed ("as we can see in the example code below"):
- `init` - to initialize key variables required by the custom AI
- `fit` - to define what kind of training the model does, depending on NN architecture or custom hardcoded logic
- `predict` - to define how the model classified pages (important to specify it classifies pages, not documents, and sets the first_page attribute and why it's necessary for flagging a split point)
- `save` - optional because the parent class already defines how to save a pickle file in a way that the Konfuzio Server can read, probably not needed to override this unless special purposes are required such as saving a json like we do here (and in that case the new implementation should make sure to call the super such that the pickle saving option is not lost)

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -85,22 +87,29 @@ def load_model(pickle_path: str, max_ram: Union[None, str] = None):
         if "unsupported pickle protocol: 5" in str(err) and '3.7' in sys.version:
             raise ValueError("Pickle saved with incompatible Python version.") from err
         raise
-
+    if not issubclass(type(model), BaseModel):
+        if (
+            "2022-03-10-15-14-51_lohnabrechnung_old_model" not in pickle_path
+            and "2022-09-27-18-45-41_lohnabrechnung" not in pickle_path
+            and "list_test" not in pickle_path
```
da-konfuzio, line 94: Can you remind me why we added specific file names hardcoded here?

konfuzio_sdk/trainer/file_splitting.py
```diff
@@ -90,35 +79,41 @@ def fit(self, *args, **kwargs) -> dict:
...
-            cloudpickle.dump(self.first_page_spans, f)
-        logger.info('Compressing model with bz2')
-        with open(temp_pkl_file_path, 'rb') as input_f:
-            with bz2.open(pkl_file_path, 'wb') as output_f:
-                shutil.copyfileobj(input_f, output_f)
-        logger.info('Deleting cloudpickle file')
-        os.remove(temp_pkl_file_path)
-        return pkl_file_path
+        if save_json:
+            path = self.output_dir + f'/{get_timestamp()}_first_page_spans.json'
```
da-konfuzio, line 97: Would it be possible to have this path variable defined in `init` or as a property?

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1324,13 +1333,110 @@ def add_extractions_as_annotations(
                 )
 
 
-class Trainer:
+class BaseModel:
+    """Base model to define a save() method for child classes."""
```
da-konfuzio, line 1337: rather to define common methods, probably we expand it in the future

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1324,13 +1333,110 @@ def add_extractions_as_annotations(
...
+        logger.info('Deleting cloudpickle file')
+        # then delete cloudpickle file
+        os.remove(temp_pkl_file_path)
+
+        size_string = f'{os.path.getsize(pkl_file_path) / 1_000_000} MB'
+
+        if not self.model_type == 'file_splitting':
+            # restore Documents of the Category so that we can run the evaluation later
+            logger.info(f'Model ({size_string}) {self.name_lower()} was saved to {pkl_file_path}')
+            self.category.project._documents = category_documents
```
da-konfuzio, line 1428: I guess you meant `if self.model_type != 'file_splitting'`? :slightly_smiling_face:

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1324,13 +1333,110 @@ def add_extractions_as_annotations(
...
+                self.output_dir = self.train_data[0].project.model_folder
+            else:
+                self.output_dir = self.category.project.model_folder
+        if include_konfuzio:
+            cloudpickle.register_pickle_by_value(konfuzio_sdk)
+            # todo register all dependencies?
+        pathlib.Path(self.output_dir).mkdir(parents=True, exist_ok=True)
+        if self.model_type == 'file_splitting':
+            temp_pkl_file_path = os.path.join(self.output_dir, f'{get_timestamp()}_{self.__class__.__name__}_tmp.pkl')
+            pkl_file_path = os.path.join(self.output_dir, f'{get_timestamp()}_{self.__class__.__name__}.pkl')
```
da-konfuzio, line 1371: Other than the class name, let's include the project ID in the pickle name

konfuzio_sdk/evaluate.py
```diff
@@ -266,6 +266,22 @@ def f1(self) -> Optional[float]:
...
+    def metrics_markdown_table_generator(self, output_dir: str = ""):
+        """
+        Create a Markdown file with metrics.
+
+        :param output_dir: A path to save the resulting Markdown file to.
+        :type output_dir: str
+        :returns: A path to the saved metrics report.
+        """
+        header = """| TP | TN | FP | FN | precision | recall | F1 |\n|----|----|----|----|----|----|----|\n"""
+        metrics = f'|{self.tp}|{self.fn}|{self.fp}|{self.fn}|{self.precision}|{self.recall}|{self.f1}|'
```
atraining, line 278: can we just log this, I know the PR is closed but the following is hard to get

- name of the file does not point to this
- logging the information would be enough without another file
- specify "metrics" 
- add documentation how to use it

atraining:
@da-konfuzio 
- link in PR to documentation is no longer working 
- tests/example_project_data/models/2022-12-28-01-58-16_lohnabrechnung.pkl and another pickle is git added whithout any need

konfuzio_sdk/trainer/file_splitting.py
```diff
@@ -51,18 +41,62 @@ def predict(self, page: Page) -> Page:
         :return: A Page with or without is_first_page label.
         """
 
+    @property
+    def temp_pkl_file_path(self):
+        """Generate a path for temporary pickle file."""
+        temp_pkl_file_path = os.path.join(
+            self.output_dir, f'{get_timestamp()}_{self.name_lower()}_{self.documents[0].project.id_}_tmp.pkl'
```
atraining, line 48: any of self.documents ... etc can be None which will raise an Attribute Error... None has no attribute id_

konfuzio_sdk/trainer/file_splitting.py
```diff
@@ -51,18 +41,62 @@ def predict(self, page: Page) -> Page:
...
+        )
+        return temp_pkl_file_path
+
+    @property
+    def pkl_file_path(self):
+        """Generate a path for a resulting pickle file."""
+        pkl_file_path = os.path.join(
+            self.output_dir, f'{get_timestamp()}_{self.name_lower()}_{self.documents[0].project.id_}.pkl'
```
atraining, line 56: see above

konfuzio_sdk/trainer/file_splitting.py
```diff
@@ -51,18 +41,62 @@ def predict(self, page: Page) -> Page:
...
+        """Remove all data not necessary for prediction."""
+        self.documents = None
+        self.test_documents = None
+
+    def reduce_model_weight(self):
+        """Remove all non-strictly necessary parameters before saving."""
+        self.lose_weight()
+        self.tokenizer.lose_weight()
+
+    def normalize_model_memory_usage(self, max_ram):
```
atraining, line 70: why not in in BaseModel defined?

konfuzio_sdk/trainer/file_splitting.py
```diff
@@ -90,35 +124,54 @@ def fit(self, *args, **kwargs) -> dict:
...
+        return self.path
+
+    def load_json(self, model_path=""):
+        """
+        Load JSON with previously gathered first_page_spans.
+
+        :param model_path: Path for the JSON.
+        :type model_path: str
+        """
+        with open(model_path, 'r') as f:
```
atraining, line 170: see data.py on how to read files from disc - you open as 'r' and don't check for file

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -86,7 +88,13 @@ def load_model(pickle_path: str, max_ram: Union[None, str] = None):
         if "unsupported pickle protocol: 5" in str(err) and '3.7' in sys.version:
             raise ValueError("Pickle saved with incompatible Python version.") from err
         raise
-
+    if not issubclass(type(model), BaseModel):
```
atraining, line 91: ?

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -86,7 +88,13 @@ def load_model(pickle_path: str, max_ram: Union[None, str] = None):
...
+        if (
+            "2022-03-10-15-14-51_lohnabrechnung_old_model" not in pickle_path
```
atraining, line 93: ?

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1200,13 +1209,100 @@ def add_extractions_as_annotations(
...
+        :param max_ram: Specify maximum memory usage condition to save model.
+        :raises MemoryError: When the size of the model in memory is greater than the maximum value.
+        :return: Path of the saved model file.
+        """
+        logger.info('Saving model')
+        logger.info(f'{include_konfuzio=}')
+        logger.info(f'{reduce_weight=}')
+        logger.info(f'{max_ram=}')
+        logger.info(f'{keep_documents=}')
+        version = pkg_resources.get_distribution("konfuzio-sdk").version
```
atraining, line 1262: test separately, add as utility to allow sdk wide usage

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1200,13 +1209,100 @@ def add_extractions_as_annotations(
...
+        logger.info(f'{version=}')
+        pathlib.Path(self.output_dir).mkdir(parents=True, exist_ok=True)
+        logger.info('Getting save paths')
+        # temp_pkl_file_path, pkl_file_path = self.generate_pickle_output_paths()
```
atraining, line 1266: lines of code can be deleted if unused

konfuzio_sdk/trainer/information_extraction.py
iftwigs: we want to allow only loading the models that were created with the usage of BaseModel as a parent class. Samuel explains this:
> We only allow new RFExtractionAI models with the SDK (by checking with issubclass), and add error handling logic in the server so that it also allows loading of models trained with the old trainer before RFExtractionAI (DocumentAnnotationMultiClassModel...) if the model doesn't inherit from BaseModel. (we don't really care about RFExtractionAI models that don't inherit from the new "BaseModel" abstract class since there are no production models in that situation.)

konfuzio_sdk/trainer/information_extraction.py
iftwigs: a hotfix until these files have been regenerated with the usage of BaseModel, Samuel is currently working on this. I'll add a comment for clarification.

PR #190, opened 2022-12-16
Added: Additional API v3 guides

mexicat:


docs/web/api-v3.md
```diff
@@ -610,7 +610,110 @@ create your annotations programmatically.
 
 ### Create training data and train the AI
 
-Coming soon.
+Once you have uploaded enough documents and created enough annotations, you can start training an
```
atraining, line 613: capitalize concepts please.

docs/web/api-v3.md
```diff
@@ -610,7 +610,110 @@ create your annotations programmatically.
...
+[extraction AI](https://help.konfuzio.com/modules/extractions/index.html). You will need at least one document in the
```
atraining, line 614: same here, Extraction AI / Document etc

docs/web/api-v3.md
```diff
@@ -610,7 +610,110 @@ create your annotations programmatically.
...
+```
+curl --request POST \
+  --url https://app.konfuzio.com/api/v3/extraction-ais/ \
+  --header 'Content-Type: application/json' \
+  --header 'Authorization: Token YOUR_TOKEN' \
+  --data '{"category": CATEGORY_ID}'
+```
+
+The training of the AI can take a while, depending on current server load and how large the training dataset is. You
+will receive an email once the process is completed. The newly trained extraction AI will then automatically be used to
```
atraining, line 631: isn't there a better suggestion ;)

the person is using the api

docs/web/api-v3.md
```diff
@@ -610,7 +610,110 @@ create your annotations programmatically.
...
+extract machine-generated annotations from newly uploaded documents for that category.
```
atraining, line 632: slightly incorrect, the Extraction AI needs to be activated first, name the automatic way and the Patch call to do so

docs/web/api-v3.md
```diff
@@ -610,7 +610,110 @@ create your annotations programmatically.
...
+Once you have uploaded enough documents and created enough annotations, you can start training an
+[extraction AI](https://help.konfuzio.com/modules/extractions/index.html). You will need at least one document in the
+"training" dataset for the category you want to train, but more data is usually better (see our
+[improving accuracy guide](https://help.konfuzio.com/tutorials/improve-accuracy/index.html)).
+
+Then to train an AI you can simply call our
+[extraction AI create endpoint](https://app.konfuzio.com/v3/swagger/#/extraction-ais/extraction_ais_create) with the
+ID of the category the training documents belong to:
+
+```
```
atraining, line 622: do we somewhere mention that examples use Token auth, however there are other ways to authenticate? otherwise please add it and link to the authentication options

docs/web/api-v3.md
```diff
@@ -610,7 +610,110 @@ create your annotations programmatically.
...
+
+We offer a [postprocess endpoint](https://app.konfuzio.com/v3/swagger/#/documents/documents_postprocess_create)
+that allows you to change uploaded documents in three ways, which can be combined into a single API request:
+
+- _Split_: divide a document into two or more documents, with the same total number of pages. Note: you cannot join
+  documents that have been split.
+- _Rotate_: change the orientation of one or more pages in a document, in multiple of 90 degrees.
+- _Sort_: change the order of the pages in a document.
+
+The endpoint accepts a list of objects, each one representing a single output document. (If you're not using the
```
atraining, line 698: object might be arbitrary, what do you mean? list of page indexes?

docs/web/api-v3.md
```diff
@@ -610,7 +610,110 @@ create your annotations programmatically.
...
+splitting functionality, this list should only contain one document). The `pages` property you send determines the
+content of the document.
+
+### Download the OCR version of an uploaded document
+
+After uploading a document, the Konfuzio server also creates an OCR version of it with indexed and selectable text.
```
atraining, line 704: to be more precise: It's PDF document according to ?? standard PDF A/1 ?? incl. OCR text.

so this includes the JPG TO PDF; PNG TO PDF; etc. 

this section is really important as many are asking for thsi feature! and it's pretty hidden ;) so perhasp a screenshot might help

image

just to make sure the resonse of the curl get request can be downloaded without providing a file name?

docs/web/api-v3.md
mexicat: I've added the alternative way to do this (poll the API endpoint until the status changes).

docs/web/api-v3.md
mexicat: After training, we do automatically set the new AI as active if its evaluation results are better than the ones from the currently active AI. Maybe the paragraph can be reworded to reflect this?

docs/web/api-v3.md
mexicat: Yes, at the beginning of the guides section we say that all examples use token authentication.

docs/web/api-v3.md
mexicat: They are JSON objects (so, dicts). I link the postprocess endpoint documentation for a more concrete example.

docs/web/api-v3.md
mexicat: I reworded this to explicitly say that we generate a PDF, but I'm unsure about the screenshot.

I've also added two arguments to the curl command so the file is saved with the name provided by the server.

docs/web/api-v3.md
mexicat: Note from @zypriafl: just add the set active logic in the doc.

docs/web/api-v3.md
mexicat: Note from @zypriafl: current wording is fine, but we could link to the [relevant part](https://help.konfuzio.com/modules/documents/index.html#pdf-file) on help.konfuzio.com and expand it to explain what an "OCR PDF" is.

zypriafl (approved):

PR #193, opened 2022-12-21
Fixed: Validate if an RFExtractionAI can be used for extraction rather than crashing

samuel500:
This creates a method `check_is_ready_for_extraction` to validate if an RFExtractionAI can be used for extraction, and fixes a few minor issues:
- The `_meta_data` attribute is now also removed when running `Project.lose_weight`
- Saving an ExtractionAI object with `reduce_weight` set to True and `keep_documents` set to False does not remove the documents in memory. 
- The `category` can now be set in the parent `Trainer` class.
- Removes a few legacy methods.
- Removes the `pathos` dependency.

### DIFF
--- a/konfuzio_sdk/data.py
+++ b/konfuzio_sdk/data.py
@@ -2379,6 +2379,7 @@ class Document(Data):
 
         :param only_multiline_labels: Only merge if multiline Label Annotation in category training set
         """
+        logger.info("Vertical merging Annotations.")
         labels_dict = {}
         for label in self.project.labels:
             if not only_multiline_labels or label.has_multiline_annotations(categories=[self.category]):
--- a/konfuzio_sdk/data.py
+++ b/konfuzio_sdk/data.py
@@ -2880,6 +2880,7 @@ class Project(Data):
         for label in self.labels:
             label.lose_weight()
         self._documents = []
+        self._meta_data = []
         return self
 
 
--- a/konfuzio_sdk/data.py
+++ b/konfuzio_sdk/data.py
@@ -784,7 +784,7 @@ class Label(Data):
 
         search = [1, 3, 5]
         regex_to_remove_groupnames = re.compile(r'<.*?>')
-        regex_to_remove_groupnames_full = re.compile(r'\(\?:\(\?P<[^>]+>([^\)]+)\)\)')
+        regex_to_remove_groupnames_full = re.compile(r'\?P<.*?>')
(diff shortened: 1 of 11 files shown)


konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1207,93 +1207,33 @@ def __init__(self, *args, **kwargs):
         """Initialize ExtractionModel."""
         # Go through keyword arguments, and either save their values to our
         # instance, or raise an error.
+        self.category: Category = kwargs.get('category', None)
+        logger.info(f"{self.category=}")
```
zypriafl, line 1211: What is the strategy of this logger statement? Seems like they are only temporary for debugging. If they stay they need to be a full sentence

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1207,93 +1207,33 @@ def __init__(self, *args, **kwargs):
         """Initialize ExtractionModel."""
         # Go through keyword arguments, and either save their values to our
         # instance, or raise an error.
+        self.category: Category = kwargs.get('category', None)
```
zypriafl, line 1210: What is the backgroud to adding 'category'?
Why do we need it to parse from kwargs, should we make it a named argument?
Is it usually expected?

konfuzio_sdk/data.py
```diff
@@ -2879,6 +2880,7 @@ def lose_weight(self):
         for label in self.labels:
             label.lose_weight()
         self._documents = []
+        self._meta_data = []
```
zypriafl, line 2883: What is this use for?

samuel500: It's where the list of all the documents is loaded into from the Project meta-data file (`documents_meta.json5`).

konfuzio_sdk/trainer/information_extraction.py
samuel500: Is there a good template for when one is logging multiple arguments at the beginning of a method or when initialising a class? For example, when I initialise RFExtractionAI, https://github.com/konfuzio-ai/konfuzio-sdk/blob/master/konfuzio_sdk/trainer/information_extraction.py#L1951
would just adding a log at the beginning like: `logger.info("Initializing RFExtractionAI with settings:")` be enough? Or should I write a sentence for each?

PR #204, opened 2023-01-11
Fix: Vertical merge sometimes not working for loaded RFExtractionAI models

samuel500:
This introduces a fix for the vertical merge logic caused by unhandled inconsistencies between the Category loaded from server and the Category saved with the RFExtractionAI model pickle file.

### DIFF
--- a/konfuzio_sdk/data.py
+++ b/konfuzio_sdk/data.py
@@ -2371,7 +2371,7 @@ class Document(Data):
         self._annotations = None
         self._annotation_sets = None
 
-    def merge_vertical(self, only_multiline_labels=True):
+    def merge_vertical(self, labels_has_multiline_annotation: Dict, only_multiline_labels=True):
         """
         Merge Annotations with the same Label.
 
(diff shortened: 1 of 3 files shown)


konfuzio_sdk/data.py
```diff
@@ -2371,21 +2371,21 @@ def delete(self):
...
         :param only_multiline_labels: Only merge if multiline Label Annotation in category training set
         """
         logger.info("Vertical merging Annotations.")
         labels_dict = {}
-        for label in self.project.labels:
-            if not only_multiline_labels or label.has_multiline_annotations(categories=[self.category]):
-                labels_dict[label.id_local] = []
+        for label in self.category.labels:
+            if not only_multiline_labels or labels_has_multiline_annotation[label.name]:
+                labels_dict[label.name] = []
```
da-konfuzio, line 2384: Can we think of a cleaner solution? It feels wrong to pass a Dict[str, bool] when the information is retrievable from `self.project.labels[i].has_multiline_annotations`

samuel500: The difficulty here is that `Label.has_multiline_annotations` gets its information from the training data, and at inference we should assume that we don't have access to that. The alternative solution would be to store a `_has_multiline_annotations` attribute for each Label that is loaded only once at training time. It would be cleaner in one way, but also would make it less clear that this data was generated during training time.

PR #205, opened 2023-01-12
Added: Methods to create and delete Documents on Konfuzio Server, and a method to patch their metadata

samuel500:
This introduces methods to create (`Document.from_file_sync` and `Document.from_file_sync`), delete (`Document.delete`) and modify the metadata (`Document.save_meta_data`) of Konfuzio Documents.


konfuzio_sdk/data.py
```diff
@@ -1710,16 +1710,138 @@ def __init__(
         self.bbox_file_path = os.path.join(self.document_folder, "bbox.zip")
         self.bio_scheme_file_path = os.path.join(self.document_folder, "bio_scheme.txt")
 
+        self.bboxes_available: bool = (
+            True
+            if (self.is_online or self._bbox_json or is_file(self.bbox_file_path, raise_exception=False))
+            else False
+        )
+
```
da-konfuzio, line 1718: Then this is more readable by just doing the following:

```python
bbox_file_exists = is_file(self.bbox_file_path, raise_exception=False)
self.bboxes_available: bool = self.is_online or self._bbox_json or bbox_file_exists
```

:wink:

da-konfuzio (approved):

PR #219, opened 2023-02-08
update initial storage requirements for on prem users

genego-io:


docs/web/on_premises.md
```diff
@@ -187,6 +187,24 @@ RAM and one Nvidia GPU with minimum 4GB which supports at least CUDA10.1 and CUD
 trying to deploy a non-production instance, you can reduce the defaults in order to fit
 into a smaller cluster. Konfuzio can work without a GPU. The GPU is used to train and run Categorization AIs. We observe a 5x faster training and a 2x faster execution on GPU compared to CPU. Most Konfuzio Installations do not use GPUs.
 
+##### Initial Storage (HDD) Requirements
```
zypriafl, line 190: I think we can just call it storage requirements. HDD sounds misleading as like dont use SSD. In other parts of the documentaiton we call it on-premises not on-prem.

zypriafl (approved):

PR #270, opened 2023-04-13
Added: `Page.img_bytes` to load a Page image from bytes

da-konfuzio:
We add `Page.img_bytes` to allow loading a Page image from bytes rather than using Pillow to load from the Project's folder. This is useful if you need to retrieve a Page image from a blob storage, since no image path would be available.

### DIFF
--- a/konfuzio_sdk/data.py
+++ b/konfuzio_sdk/data.py
@@ -124,6 +124,7 @@ class Page(Data):
         self.copy_of_id = copy_of_id
         self.text_encoded: List[int] = None
         self.image = None
+        self.img_bytes = None
         self._original_size = original_size
         self.width = self._original_size[0]
         self.height = self._original_size[1]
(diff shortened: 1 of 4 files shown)


konfuzio_sdk/data.py
```diff
@@ -172,6 +173,11 @@ def __repr__(self):
 
     def get_image(self, update: bool = False):
         """Get Document Page as PNG."""
```
zypriafl, line 175: This seems to do the right thing, however this functions has gotten quite complex. 
- Can we describe the difference between "image" and "img_bytes" (e.g. in the docstring)
- Should we rename img_bytes to image_bytes for consistency?
-  "self.document.status[0] == 2" is the a constant value we can use instead of 2 to improve readibility? e.g.  self.document.status[0] == Document.DONE
- Can we add type hint to "def get_image(self, update: bool = False)", "image" and "img_bytes"?

zypriafl (approved):

PR #273, opened 2023-04-20
Fixed: Default Evaluation filters out the case where a Label should only appear once per AnnotationSet

da-konfuzio:
When selecting Default Evaluation: After we have calculated the TPs, FPs, FNs for the Document, we filter out the case where a Label should only appear once per AnnotationSet but has been predicted multiple times. In this case, if any of the predictions is a TP then we keep one and discard FPs/FNs. If no TPs, if any of the predictions is a FP then we keep one and discard the FNs. If no FPs, then we keep a FN. The prediction we keep is always the first in terms of start_offset.

To see some examples of how it works in practice, see [Konfuzio Guide - Extraction AI Evaluation](https://help.konfuzio.com/modules/extractions/index.html#evaluation).

### DIFF
--- a/konfuzio_sdk/evaluate.py
+++ b/konfuzio_sdk/evaluate.py
@@ -39,6 +39,7 @@ RELEVANT_FOR_EVALUATION = [
     "end_offset_predicted",
     # "is_correct_predicted", # it's a prediction so we don't know if it is correct
     "label_id_predicted",
+    "label_has_multiple_top_candidates_predicted",
     "label_threshold_predicted",  # we keep a flexibility to be able to predict the threshold
     # "revised_predicted",  # it's a prediction so we ignore if it is revised
     "annotation_set_id_predicted",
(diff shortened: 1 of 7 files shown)


konfuzio_sdk/evaluate.py
```diff
@@ -101,7 +102,9 @@ def compare(
     df_a = pandas.DataFrame(doc_a.eval_dict(use_correct=only_use_correct))
     df_b = pandas.DataFrame(
         doc_b.eval_dict(
-            use_view_annotations=use_view_annotations, use_correct=False, ignore_below_threshold=ignore_below_threshold
+            use_view_annotations=strict and use_view_annotations,  # view_annotations only available for strict mode
```
zypriafl, line 105: "view_annotations only available for strict mode"

Why is that the case? Can you add a short explanation as comment?

da-konfuzio: Thanks I added explanation to the docstring

zypriafl (approved):

PR #285, opened 2023-05-03
Added: A method to get a Page by ID from the Document, a method to get a Page's original from its copy

iftwigs:
Added methods to retrieve an original Page from its copy and to get a Page by its ID from the Document; relaxed the conditions of setting Category – now a Document's Category will always be NO_CATEGORY when specified as None.


zypriafl (dismissed):

tests/trainer/test_file_splitting.py
```diff
@@ -209,27 +214,26 @@ def test_splitting_no_category_document(self):
...
-        test_document = self.file_splitting_model.tokenizer.tokenize(deepcopy(self.test_document))
-        pred = splitting_ai.propose_split_documents(test_document, return_pages=True)[0]
+        pred = splitting_ai.propose_split_documents(self.test_document, return_pages=True)[0]
         for page in pred.pages():
             if page.number in (1, 3, 5):
                 assert page.is_first_page
             else:
                 assert not page.is_first_page
             assert page.is_first_page_confidence == 1
+        assert self.test_document.pages()[0].category == pred.pages()[0].category
```
zypriafl, line 236: Why do we the test only the first page ("[0]")?

tests/trainer/test_file_splitting.py
```diff
@@ -209,27 +214,26 @@ def test_splitting_no_category_document(self):
...
             assert page.is_first_page_confidence == 1
-        assert pred == test_document
+        assert pred == self.test_document
 
     def test_suggest_first_pages(self):
         """Test Splitting AI's suggesting first Pages."""
         splitting_ai = SplittingAI(self.file_splitting_model)
-        test_document = self.file_splitting_model.tokenizer.tokenize(deepcopy(self.test_document))
-        pred = splitting_ai.propose_split_documents(test_document, return_pages=True)[0]
+        pred = splitting_ai.propose_split_documents(self.test_document, return_pages=True)[0]
```
zypriafl, line 229: We could improve reaibility by chaning this line to:

```
predictions = splitting_ai.propose_split_documents(self.test_document, return_pages=True)
assert len(predictions) == 1
predicition = predictions[0] 
        
   ``

File: tests/trainer/test_file_splitting.py [hunk 3]
```diff
@@ -187,6 +187,11 @@ def test_splitting_ai_evaluate_full_on_training(self):
 
     def test_splitting_ai_evaluate_full_on_testing(self):
         """Test Splitting AI's evaluate_full on testing Documents."""
+        counter = 1
+        for doc in self.file_splitting_model.test_documents:
```
zypriafl, line 191: Can you add a comment explaining, why we need to set all those IDs in this test case here?

konfuzio_sdk/data.py
```diff
@@ -435,6 +436,23 @@ def category(self) -> Optional['Category']:
         else:
             return self._category
 
+    def get_original_page(self) -> 'Page':
+        """
+        Return an original Page in case the current Page is a copy without an ID.
```
zypriafl, line 441: Please explain what "original" means here. Why do we need to get the original Page instance. Was ist different here? (=> please add this to the doctring)

docs/sdk/SDK and Server Integration.drawio
zypriafl, line 1: Unfortunately, I cannot see the changes compared to the last version. I see a lot of "upcoming features" comments... I think they need to be updated.

image

tests/trainer/test_file_splitting.py
iftwigs: changed to checking all the pages

docs/sdk/SDK and Server Integration.drawio
iftwigs: updated them

zypriafl (approved):

PR #292, opened 2023-05-15
Fix: get_segmentation call from Virtual Document

samuel500:
This fixes a bug in the recently introduced `get_segmentation` method which caused the segmentation API to be called when the method was called with a  Virtual Document even when the segmentation data is already available in the original Document.


konfuzio_sdk/api.py
```diff
@@ -134,17 +134,22 @@ def build_response(self, req, resp):
         return response
 
 
-def konfuzio_session(token: str = KONFUZIO_TOKEN, timeout: Optional[int] = None):
+def konfuzio_session(token: str = KONFUZIO_TOKEN, timeout: Optional[int] = None, num_retries: Optional[int] = None):
```
zypriafl, line 137: Why not "Optional[int] = 5" ?

How did we come up with 5? Do we use it somewhere else?

samuel500: About the default values, I'm not sure why they were chosen originally. 

As to why not `"Optional[int] = 5" `, the intention is that when we pass a custom timeout here: https://github.com/konfuzio-ai/konfuzio-sdk/blob/11029-get-segmentation-fix/konfuzio_sdk/data.py#L2600 or https://github.com/konfuzio-ai/konfuzio-sdk/blob/11029-get-segmentation-fix/konfuzio_sdk/data.py#L2732 ,  we can just pass None and it still takes on the default value. The alternative would be to duplicate the definition of the default timeout values in from_file and get_segmentation.

zypriafl (approved):

PR #306, opened 2023-05-25
Changed: Default input parameters for the Categorization AI's fitting

iftwigs:
Changed the format accepted as an input at Categorization AI's `fit()` method; set default values to the `batch_size` and `max_len` arguments.

### DIFF
--- a/konfuzio_sdk/trainer/document_categorization.py
+++ b/konfuzio_sdk/trainer/document_categorization.py
@@ -359,7 +359,7 @@ class NBOWSelfAttention(AbstractTextCategorizationModel):
     def _load_architecture(self) -> None:
         """Load NN architecture."""
         self.embedding = nn.Embedding(self.input_dim, self.emb_dim)
-        self.multihead_attention = nn.MultiheadAttention(self.emb_dim, self.n_heads, batch_first=True)
+        self.multihead_attention = nn.MultiheadAttention(self.emb_dim, self.n_heads)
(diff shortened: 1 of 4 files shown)


tests/trainer/test_document_categorization.py
```diff
@@ -556,15 +551,16 @@ def test_build_categorization_ai() -> None:
     project = Project(id_=None, project_folder=OFFLINE_PROJECT)
     categorization_pipeline = build_categorization_ai_pipeline(
         categories=project.categories,
-        documents=project.documents,
-        test_documents=project.test_documents,
+        documents=project.documents[:5],
```
zypriafl, line 554: Can we make this even faster by just using one Document?

konfuzio_sdk/trainer/document_categorization.py
```diff
@@ -1456,6 +1454,8 @@ def build_categorization_ai_pipeline(
     categorization_pipeline.category_vocab = categorization_pipeline.build_template_category_vocab()
     # Configure image and text models
     if image_model is not None:
+        if isinstance(image_model, str):
```
zypriafl, line 1457: Can we simplify this using "ImageModel.__members__.get(string)" ?
Can we raise an exception if there is no match

iftwigs: added the exceptions. 
it seems there's no way of getting the element by the value with the enum; I can remake it into a simpler data structure or leave it as it is now.

PR #309, opened 2023-05-30
Changed: Code examples in the Documentation contain relative references to the .py scripts

iftwigs:


iftwigs:
@zypriafl the pickle was changed because the previous version (used in code example for document categorization testing) didn't have keep_documents=True upon saving, so it wasn't possible to use for testing of the evaluation of the loaded model, which is one of the parts of the code example.

PR #313, opened 2023-06-02
Fixed: File Splitting Evaluation's ZeroDivisionError handling

iftwigs:
Fixed handling ZeroDivisionErrors in File Splitting Evaluation when no True Positives were predicted and values of precision, recall and F1 have not been calculated.


zypriafl:
Hi Lisa,

I would suggest to adapt the behviour of sklear.
https://scikit-learn.org/stable/modules/generated/sklearn.metrics.precision_score.html

They use zero_division argument which acts as defualt.

iftwigs:
@zypriafl updated the PR with the approach suggested by you.

zypriafl (approved):

PR #314, opened 2023-06-06
Changed: BERT and image processing architecture used for Multimodal File Splitting model; lz4 compression for pickle files

iftwigs:
* Replaced legal-bert-base-uncased with legal-bert-small-uncased to speed up the loading and minimize the size; removed several convolution layers from the image processing module.
* lz4 compression replaces bz2 by default.


setup.py
```diff
@@ -76,6 +76,7 @@
     install_requires=[
         'cloudpickle==2.2.1',  # Used to pickle objects
         'filetype==1.0.7',  # Used to check that files are in the correct format
+        'lz4<5',  # Used to compress pickles
```
zypriafl, line 79: Why do we need to restrict to "<5".

zypriafl: then lets remove it here. the SDK is often used as library where uses add additional libraries. we should not forbid to use the latest version of a package if there is no reason.

zypriafl (approved):

PR #321, opened 2023-06-20
Added: Use of relative position and first word features in Extraction AI

samuel500:
This adds positional features relative to the Page and additional string features relating to the first word of a Page to the features used by the RFExtractionAI.


konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -870,7 +870,13 @@ def process_document_data(
     df["relative_position_in_page"] = df["page_index"] / document_n_pages
 
     abs_pos_feature_list = ["x0", "y0", "x1", "y1", "page_index", "area_quadrant_two", "area"]
-    relative_pos_feature_list = ["relative_position_in_page"]
+    relative_pos_feature_list = [
```
zypriafl, line 873: Can we add a testcase which assert the feature names (i.e. column names) in the feature dataframe?

zypriafl (approved):

PR #336, opened 2023-07-10
Fix: Regex precision calculation bug

samuel500:
This fixes an issue in the recall calculation which caused the recall value to be greater than 1.


zypriafl (commented):
The MR looks good. Only one question, was there no existing test for actual f1-score calculation? Could we add one test that replicates the issue that f1-score was over 1.

PR #337, opened 2023-07-10
Changed:  Error message about missing dependencies

iftwigs:
Changed the message about missing AI-related dependencies to a more user-friendly one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,9 +56,7 @@ By default, the SDK is installed without the AI-related dependencies like `torch
 only the Data-related SDK concepts but not the AI models. To install the SDK with the AI components,
 run the following command:
   ```
-  git clone https://github.com/konfuzio-ai/konfuzio-sdk.git
-  cd konfuzio-sdk
-  pip install -e .[ai]
+  pip install konfuzio_sdk[ai]
   ```
 
 Find the full installation guide [here](https://dev.konfuzio.com/sdk/get_started.html#install-sdk)
--- a/konfuzio_sdk/settings_importer.py
+++ b/konfuzio_sdk/settings_importer.py
@@ -32,7 +32,11 @@ def is_dependency_installed(dependency: str) -> bool:
     return dependency in {pkg.key for pkg in pkg_resources.working_set}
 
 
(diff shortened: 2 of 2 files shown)


docs/sdk/tutorials/ner-ontonotes-fast/index.md
```diff
@@ -0,0 +1,14 @@
+Retrain Flair NER-Ontonotes-Fast with Human Revised Annotations
+---------------------------------------------------------------
+
+The tutorial *HRetrain Flair NER-Ontonotes-Fast with Human Revised Annotations* aims to show you how to use the
```
zypriafl, line 4: This seems to be a typo? "HRetrain"

extras_list.py
```diff
@@ -2,11 +2,11 @@
 
 EXTRAS = {
     'dev': [
-        'flake8',
-        'pydocstyle',
-        'pytest',
-        'pre-commit',
-        'parameterized',
+        'flake8==6.0.0',
```
zypriafl, line 5: Why do we need to be that strict?

setup.py
```diff
@@ -78,19 +78,19 @@
...
         'matplotlib==3.7.1',
-        'nltk',
+        'nltk==3.6.3',
         'numpy>=1.22.4',
         'pandas>=1.3.5,<2.0.0',
         'Pillow>=8.4.0',
-        'python-dateutil',
-        'python-decouple',  # todo add ==3.3 ?
-        'requests',  # todo add ==2.24.0 ?
+        'python-dateutil==2.8.2',
```
zypriafl, line 87: Why do we need to be that strict?

iftwigs: Chris suggested we minimize the number of libraries with an unspecified version, because Mohamed experienced some problems during setting up the SDK environment because of that. already having had some packages installed caused him to have conflicting dependencies

zypriafl: Can we make them ">=" because if a Uset wants to include the SDK in his python project, we would forbid him to use the latest version of a package as soon a package brings out a new version.

zypriafl (approved):

PR #340, opened 2023-07-11
Fix: Bug in evaluation DataFrame where false positives are not associated with any Document

samuel500:
This fixes an issue where false positive rows in the evaluation DataFrame had no Document id associated with the row.


tests/test_evaluate.py
```diff
@@ -214,6 +214,19 @@ def test_strict_nothing_should_be_predicted(self):
         assert evaluation["tokenizer_false_positive"].sum() == 24
         assert evaluation["tokenizer_false_negative"].sum() == 0
 
+    def test_false_positive_document_id_predicted(self):
```
zypriafl, line 217: Why did we do a new test case here, could we not reuse: https://github.com/konfuzio-ai/konfuzio-sdk/blob/master/tests/test_evaluate.py#L201 for example?

samuel500: Wanted to created an isolated test just for this, but you're right, it's not really necessary.

zypriafl (approved):

zypriafl (approved):

PR #343, opened 2023-07-12
Fix: Bug preventing saving of custom Extraction AI with no Tokenizer

samuel500:
This fixes a bug where Custom Extraction AIs without Tokenizer were prevented from being saved.

### DIFF
--- a/tests/trainer/test_information_extraction.py
+++ b/tests/trainer/test_information_extraction.py
@@ -19,7 +19,7 @@ import pandas as pd
 from sklearn.datasets import make_classification
 from sklearn.ensemble import RandomForestClassifier
 
-from konfuzio_sdk.data import Project, Document, AnnotationSet, Annotation, Span, LabelSet
+from konfuzio_sdk.data import Category, Project, Document, AnnotationSet, Annotation, Span, LabelSet
 
 from konfuzio_sdk.api import upload_ai_model
 from konfuzio_sdk.settings_importer import is_dependency_installed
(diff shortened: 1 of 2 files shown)


tests/trainer/test_information_extraction.py
```diff
@@ -1362,6 +1361,54 @@ def __init__(self):
...
+        assert len(empty_doc.annotations(use_correct=False)) == 0
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        """Clear created files."""
+        is_file(cls.pipeline.pipeline_path, raise_exception=True)
+
+        os.remove(cls.pipeline.pipeline_path)
+
+
```
iftwigs, line 1411: why do we need the first class, just for the speeding the test process up?

samuel500: The first class is just the definition of the CustomAI class that we then initialize and test in the second unittest class. Would you do it differently?

iftwigs: my understanding was that basically any AI that derives from the AbstractExtractionAI is a custom Extraction AI purely because of structure, and we have tests for those. I guess we can leave this for the sake of having an example of how a minimal test setup can look for a custom Extraction AI.

iftwigs (approved):

tests/trainer/test_information_extraction.py
iftwigs: got you, agreed. I approved.

PR #345, opened 2023-07-14
Fix: Uploading, updating and deleting the AI models

iftwigs:
Fixed uploading, updating and deleting methods for the AI models via the call to the API v.3.


konfuzio_sdk/api.py
```diff
@@ -563,15 +563,21 @@ def get_results_from_segmentation(doc_id: int, project_id: int, session=konfuzio
...
-def upload_ai_model(ai_model_path: str, category_ids: List[int] = None, session=konfuzio_session()):  # noqa: F821
+def upload_ai_model(
+    ai_model_path: str, project_id: int = None, category_id: int = None, session=konfuzio_session()
+):  # noqa: F821
     """
     Upload an ai_model to the text-annotation server.
 
     :param ai_model_path: Path to the ai_model
-    :param category_ids: define ids of Categories the model should become available after upload.
+    :param project_id: An ID of a Project to which the AI is uploaded. Needed for the File Splitting and Categorization
```
zypriafl, line 573: Where do we explain the user the concept that Extraction AI belongs to a Category and Categorization AI and FileSplitting to a Project?

konfuzio_sdk/api.py
```diff
@@ -634,24 +638,19 @@ def update_ai_model(ai_model_id: int, ai_type: str, session=konfuzio_session(),
     following: 'file_splitting', 'extraction', 'categorization'.
     :param session: session to connect to the server.
     """
-    if ai_type not in ['file_splitting', 'extraction', 'categorization']:
-        raise ValueError("ai_type should have the value of 'file_splitting', 'extraction', or 'categorization'.")
+    if ai_type not in ['filesplitting', 'extraction', 'categorization']:
```
zypriafl, line 641: Please change file_splitting => filesplitting also in the comment.

konfuzio_sdk/api.py
```diff
@@ -613,8 +617,8 @@ def delete_ai_model(ai_model_id: int, ai_type: str, session=konfuzio_session()):
...
+        raise ValueError("ai_type should have the value of 'filesplitting', 'extraction', or 'categorization'.")
```
zypriafl, line 621: Should we create constants for 'filesplitting', 'extraction', or 'categorization'?

konfuzio_sdk/api.py
```diff
@@ -587,19 +593,17 @@ def upload_ai_model(ai_model_path: str, category_ids: List[int] = None, session=
...
         with open(ai_model_path, 'rb') as f:
-            multipart_form_data = {'ai_model': (model_name, f)}
+            multipart_form_data = {'file': (model_name, f)}
             headers = {"Prefer": "respond-async"}
-            r = session.post(url, files=multipart_form_data, headers=headers)
+            if not ai_type == 'extraction':
+                data = {'project': str(project_id)}
+            else:
+                data = {'category': str(category_id)}
+            r = session.post(url, files=multipart_form_data, data=data, headers=headers)
```
zypriafl, line 602: I think we need to check return codes. Is there a general guideline in the api.py how to check return codes.
In swagger we should see the expected status.

zypriafl: We might want to use raise_for_status() as it gives an explanatory error message.

zypriafl (commented):
We were only able to look at "konfuzio_sdk/api.py" however, I assume to comments are already helpful.

zypriafl (dismissed):

zypriafl (dismissed):

PR #347, opened 2023-07-18
Added: Saving Categorization AI models with extension compatible to the Server

iftwigs:
Categorization AI models are now saved as .pt.lz4, which enables them to be uploaded to the Server.


konfuzio_sdk/trainer/base.py
```diff
@@ -84,7 +84,7 @@ def load_model(pickle_path: str, max_ram: Union[None, str] = None):
         prev_local_id = next(Data.id_iter)
 
         try:
-            if pickle_path.endswith(".pt"):
+            if 'pt' in pickle_path.split('.'):
```
zypriafl, line 87: What would happen with something like "pt.zip".

konfuzio_sdk/trainer/document_categorization.py
```diff
@@ -948,7 +951,7 @@ def __init__(
         self.device = torch.device('cuda' if (torch.cuda.is_available() and use_cuda) else 'cpu')
         self.train_transforms = None
 
-    def save(self, path: Union[None, str] = None, reduce_weight: bool = True) -> str:
+    def save(self, path: Union[None, str] = None, tmp_path: Union[None, str] = None, reduce_weight: bool = True) -> str:
```
zypriafl, line 954: Why do have "tmp_path"? why is needed? If yes can we explain it in the docstring.

mexicat, line 954: I'm wondering if we could make these parameters consistent with `BaseModel`, where instead of `path` and `tmp_path` we just have `output_dir`. Otherwise the server has to generate the filename (both normal and tmp) on its side, as it can't rely on the default one from the SDK where the directory is not writable. If we had just `output_dir` instead, we could just pass a temporary path and have the SDK generate the filenames, like we do with Extraction AIs.

mexicat: If we do this, maybe it's worth adding a check for the old parameters in the function that returns an error saying "this parameter is deprecated, use `output_dir` instead".

mexicat (dismissed):

konfuzio_sdk/trainer/document_categorization.py
```diff
@@ -1657,18 +1707,28 @@ def load_categorization_model(pt_path: str, device: Optional[str] = 'cpu'):
...
-            device = 'cuda'
-        else:
-            device = 'cpu'
+        device = 'cuda' if torch.cuda.is_available() else 'cpu'
+
+    if pt_path.endswith('lz4'):
+        with open(pt_path, 'rb') as f:
+            compressed = f.read()
+        decompressed_data = lz4.frame.decompress(compressed)
+        file_data = torch.load(io.BytesIO(decompressed_data), map_location=torch.device(device))
```
zypriafl, line 1716: @iftwigs This looks like the AI can only be used on the device setting where it was saved? Is this the case? Is this documented?

konfuzio_sdk/trainer/base.py
```diff
@@ -247,6 +247,8 @@ def save(
 
         if hasattr(self, 'remove_dependencies'):
             self.remove_dependencies()
+        if hasattr(self, 'client'):
```
zypriafl, line 250: Why is this needed please add a comment?

zypriafl (approved):

PR #356, opened 2023-07-28
Fix: Bbox Class Method from_image_size()

MohamedAmineDHIAB:
- I updated the `.gitignore` to include the following :

```
dist #due to setup from source
build #due to setup from source
.vscode #vscode config
tests/example_project_data #it was updated when I ran the tests
```
- There is no typo on the `from_image_size()` method ( check the comments on [this issue](https://git.konfuzio.com/konfuzio/objectives/-/issues/11448) ). Instead, I expanded the description of the method to be more intuitive:
```diff
+         Compute the coordinates of the bottom-left and the top-right corners in
+         a (down-to-up y-axis, left-right x-axis, page-scaled system).

+        :param: (x0,y0) and (x1,y1) are the coordniates of resp. the top-left and the bottom-right corners
+        in a (up-to-down y-axis, left-right x-axis, image-scaled) system.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ dist
 build
 .vscode
 tests/example_project_data
+__pycache__
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,7 @@ settings.py
 *.tar.gz
 *data_*
 *.ipynb*
+dist
+build
+.vscode
+tests/example_project_data
--- a/konfuzio_sdk/data.py
+++ b/konfuzio_sdk/data.py
@@ -262,7 +262,7 @@ class Page(Data):
     def text(self):
         """Get Document text corresponding to the Page."""
         doc_text = self.document.text
-        page_text = self.document.text[self.start_offset : self.end_offset]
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -11,3 +11,8 @@ settings.py
 *.tar.gz
 *data_*
 *.ipynb*
+dist
+build
+.vscode
+tests/example_project_data
```
iftwigs, line 17: why do we exclude this? we need this data for testing.

konfuzio_sdk/data.py
```diff
@@ -615,8 +618,8 @@ def from_image_size(cls, x0, x1, y0, y1, page: Page) -> 'Bbox':
 
         temp_y0 = (image_height - y0) * factor_y
         temp_y1 = (image_height - y1) * factor_y
-        y0 = temp_y1
-        y1 = temp_y0
+        y0 = temp_y0
+        y1 = temp_y1
         x0 = x0 * factor_x
         x1 = x1 * factor_x
```
iftwigs, line 624: can we add a test for this method?

iftwigs (approved):

PR #361, opened 2023-08-09
update getting started docs: sdk on Apple M

tizianococcio:
Although, probably once [this](https://github.com/konfuzio-ai/konfuzio-sdk/pull/352) is done there won't be any need to tweak `setup.py`.


docs/sdk/configuration_reference.md
```diff
@@ -19,6 +19,80 @@ run the following command:
...
+
+Create a file called `.env` in the root of the Deepnote project. We will later use this file to store the credentials we obtained via the Colab notebook:
+<p align="center">
+  <img src="img/deepnote-newfile.png" width=350 alt="Deepnote create file">
+</p>
+
+To install the SDK in Deepnote, run the following commands in a new Notebook cell:
+```
+!git clone https://github.com/konfuzio-ai/konfuzio-sdk.git
+!sed -i 's/torch>=1\.8\.1/torch==1.8.1/' konfuzio-sdk/extras_list.py
```
MohamedAmineDHIAB, line 47: I think it is not a good idea to show to the user fixing versions manually, maybe creating another pull request to fix the version is better

iftwigs: We don't need to change the version, this range was set up intentionally because torch==1.8.1 will not work with some of python versions older than 3.8. We can just remove this part completely.

MohamedAmineDHIAB (approved):

PR #404, opened 2023-10-09
Added: Test coverage reports

iftwigs:
Added the automatic adding of coverage report in comments of the PR updated with every push.


konfuzio_sdk/data.py
```diff
@@ -4420,3 +4421,7 @@ def export_ais(project: Project) -> None:
 
                 print(f"[SUCCESS] exported AI model to {file_name}")
 
+
+def method_for_coverage_testing():
```
zypriafl, line 4425: What is this used for, Can you add more detailed explanation?

iftwigs: removed it, it was for testing one of now-discarded coverage testing actions that focused only on the code that has been changed inside a current PR.

PR #407, opened 2023-10-11
Fixed: Optimize spans/annotations data structures

mexicat:
* Change data structure for spans and annotations from list to dict
* Replace naive `in` checks with `dict.get`
* Ensure external interface (non-private functions) stay consistent
* Fix tests that checked private properties directly

### DIFF
--- a/konfuzio_sdk/data.py
+++ b/konfuzio_sdk/data.py
@@ -2109,7 +2109,7 @@ class Annotation(Data):
         else:
             self.custom_offset_string = None
         self.id_ = id_  # Annotations can have None id_, if they are not saved online and are only available locally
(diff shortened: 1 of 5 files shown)


konfuzio_sdk/data.py
```diff
@@ -2109,7 +2109,7 @@ def __init__(
         else:
             self.custom_offset_string = None
         self.id_ = id_  # Annotations can have None id_, if they are not saved online and are only available locally
-        self._spans: List[Span] = []
+        self._spans: Dict = {}
```
zypriafl, line 2112: Can we somewhere explain the datastructure of this dict?

zypriafl (approved):

PR #421, opened 2023-10-30
Changed: Tutorials section of documentation

iftwigs:
Changed the structure and the contents of tutorials section of documentation.


docs/sdk/tutorials/outlier-annotations/index_nb.md
```diff
@@ -0,0 +1,147 @@
...
+
+project = Project(id_=YOUR_PROJECT_ID)
+label = project.get_label_by_name(YOUR_LABEL_NAME)
+```
+
+```python editable=true slideshow={"slide_type": ""} tags=["remove-cell"]
+pipeline = RFExtractionAI()
+pipeline.tokenizer = ListTokenizer(tokenizers=[])
+pipeline.category = label.project.get_category_by_id(id_=63)
+train_doc_ids = {44823, 44834, 44839, 44840, 44841}
```
mexicat, line 77: Here you have to add the category ID and train doc IDs yourself, which is not explained. I would either comment this or find a way to derive these IDs directly - I guess you chose to hardcode them because there are too many training docs in project 46, but maybe it could just be a slice of `pipeline.category.documents()`? Just to make it easy for someone to try it without having to go fetch these IDs.

docs/sdk/tutorials/outlier-annotations/index_nb.md
```diff
@@ -0,0 +1,147 @@
...
+project = Project(id_=YOUR_PROJECT_ID)
+
+label = project.get_label_by_name(YOUR_LABEL_NAME)
+
+# get outliers by regex
+outliers = label.get_probable_outliers_by_regex(project.categories, top_worst_percentage=TOP_WORST)
+
+# get outliers by confidence
+evaluation = ExtractionEvaluation(documents=list(zip(pipeline.documents, predictions)), strict=False)
+outliers = label.get_probable_outliers_by_confidence(evaluation, confidence=CONFIDENCE)
```
mexicat, line 137: This code will not run unless you followed the steps in the other paragraph before, as `pipeline` and `CONFIDENCE` will not be defined otherwise, so the part where it says "Below is the full code to accomplish this task" is not actually true. It's probably enough to add two lines where these two variables get declared.

docs/sdk/tutorials/outlier-annotations/index_nb.md
iftwigs: this is a hidden cell (there's a tag "remove-cell"), so it is not going to be visible for the user. we need it to have a trained instance of an `RFExtractionAI` to later create an instance of `EvaluationExtraction` over its results and use the confidence predictions from the evaluation class as the input for the method `get_probable_outliers_by_confidence`.

docs/sdk/tutorials/outlier-annotations/index_nb.md
iftwigs: you're right. I will add lines with placeholders of the missing components and mention that they need to be declared with real values in order for this to run.

zypriafl (approved):

PR #431, opened 2023-11-13
Added: Usage of AutoTokenizer with BERT for Document Categorization

iftwigs:
If BERT model is selected for text-based Document Categorization, tokenization will be run using AutoTokenizer.


konfuzio_sdk/trainer/document_categorization.py
```diff
@@ -1179,13 +1179,17 @@ def build_document_classifier_iterator(
...
                 if use_text:
-                    # if using a text module, tokenize the page, trim to max length and then numericalize
-                    # REPLACE page_tokens = tokenizer.get_tokens(page_text)[:max_len]
-                    # page_encoded = [text_vocab.stoi(span.offset_string) for span in
-                    # self.spans(start_offset=page.start_offset, end_offset=page.end_offset)]
-                    # document_tokens.append(torch.LongTensor(page_encoded))
-                    self.text_vocab.numericalize(page)
-                    document_tokens.append(torch.LongTensor(page.text_encoded))
+                    if self.classifier.text_model.__class__.__name__ == 'bert':
+                        self.tokenizer = transformers.AutoTokenizer.from_pretrained(self.classifier.text_model.name)
```
MohamedAmineDHIAB, line 1183: I think we need to set `truncation=True, padding="max_length"`

tests/trainer/test_document_categorization.py
```diff
@@ -432,17 +432,17 @@ def test_forward(self) -> None:
         (WhitespaceTokenizer, LSTM, VGG, 'vgg13'),
         (ConnectedTextTokenizer, NBOW, VGG, 'vgg11'),
         (ConnectedTextTokenizer, LSTM, VGG, 'vgg13'),
-        # (PhraseMatcherTokenizer, BERT, VGG, 'vgg16'),  # PhraseMatcherTokenizer is WIP and not available for usage yet
+        (transformers.AutoTokenizer, BERT, VGG, 'vgg16'),
```
MohamedAmineDHIAB, line 435: make input_dim=None and check if it's set automatically (preferably we want it to be set automatically)

tests/trainer/test_document_categorization.py
```diff
@@ -468,13 +468,13 @@ def tearDownClass(cls) -> None:
 
     def test_1a_configure_dataset(self) -> None:
```
MohamedAmineDHIAB, line 469: * On `test_1b_configure_pipeline` make sure you load the tokenizer for BERT with `from_pretrained` and also set `truncation=True, padding="max_length"` for the tokenizer params
* Check that saved "pickles" for categorization AI get loaded successfully
* use `"prajjwal1/bert-tiny"` as `model_name` in the tests instead of `'bert-base-german-dbmdz-uncased' `to make tests faster and also cover more models

konfuzio_sdk/trainer/document_categorization.py
```diff
@@ -1180,9 +1180,13 @@ def build_document_classifier_iterator(
                     # so we just have a list of None to keep the lists the same length
                     document_images.append(None)
                 if use_text:
-                    if self.classifier.text_model.__class__.__name__ == 'bert':
+                    if self.classifier.text_model.__class__.__name__ == 'BERT':
                         self.tokenizer = transformers.AutoTokenizer.from_pretrained(self.classifier.text_model.name)
-                        document_tokens = self.tokenizer(document.text, truncation=True, padding=max_len)
+                        document_tokens.append(
```
MohamedAmineDHIAB, line 1185: I think the value of the key 'input_ids' in the output of the tokenizer is already a float tensor. in `TextualSplittingAI` I directly feed it to the model without the need for `LongTensor` transformation. Do you know why we need the LongTensor here or if tests are failing due to that ?

konfuzio_sdk/trainer/document_categorization.py
```diff
@@ -1216,7 +1220,14 @@ def collate(batch, transforms) -> Dict[str, LongTensor]:
                 image = None
             if use_text:
                 # if we are using text, batch and pad the already tokenized and numericalized text and place on GPU
-                text = torch.nn.utils.rnn.pad_sequence(text, batch_first=True, padding_value=self.text_vocab.pad_idx)
+                if not isinstance(self.tokenizer, transformers.BertTokenizerFast):
+                    text = torch.nn.utils.rnn.pad_sequence(
+                        text, batch_first=True, padding_value=self.text_vocab.pad_idx
+                    )
+                else:
+                    text = torch.nn.utils.rnn.pad_sequence(
```
MohamedAmineDHIAB, line 1228: Two questions here:
1- Why are we padding here ? I thought the tokenizer automatically does the padding and the truncation
2- If we should do the padding here and no other solution is found (although the idea is that if we use AutoTokenizer we try to remove any extra code for text pre-processing) , then in this case the padding_value for `transformers` models can be found in `model.config.pad_token_id` (I think one needs to check if for every bert model it's like that)

tests/trainer/test_document_categorization.py
```diff
@@ -488,28 +488,30 @@ def test_1b_configure_pipeline(self) -> None:
         if self.text_class is not None:
             if self.tokenizer == PhraseMatcherTokenizer:
                 self.categorization_pipeline.tokenizer = self.tokenizer(self.categorization_pipeline.documents)
-            elif self.tokenizer == transformers.AutoTokenizer:
+                self.categorization_pipeline.text_vocab = self.categorization_pipeline.build_text_vocab()
+            elif self.tokenizer == transformers.BertTokenizerFast:
                 self.categorization_pipeline.tokenizer = transformers.AutoTokenizer.from_pretrained(
-                    self.text_class.name
+                    'bert-base-german-cased'
```
MohamedAmineDHIAB, line 494: don't you think this should be defined as a variable ? or should we leave it as `self.text_class.name` and test the different names in the tests specific for different BERT-based models.

tests/trainer/test_document_categorization.py
```diff
@@ -488,28 +488,30 @@ def test_1b_configure_pipeline(self) -> None:
...
         self.categorization_pipeline.category_vocab = self.categorization_pipeline.build_template_category_vocab()
 
         image_model = None
         text_model = None
         if self.image_class is not None:
             image_model = self.image_class(name=self.image_nn_version)
         if self.text_class is not None:
-            text_model = self.text_class(input_dim=len(self.categorization_pipeline.text_vocab))
+            if self.tokenizer == transformers.BertTokenizerFast:
+                text_model = self.text_class(input_dim=512)
```
MohamedAmineDHIAB, line 507: input_dim should also be dynamic, maybe use `max_position_embeddings` , although it doesn't matter now since we will just delete it when refactoring the categorization code

konfuzio_sdk/trainer/document_categorization.py
iftwigs: before this transformation, the tests and the fitting pipeline have been failing. we can see that between lines [1197](https://github.com/konfuzio-ai/konfuzio-sdk/blob/d3ec71972df9a7eb507c2d4636bf4ef30052586e/konfuzio_sdk/trainer/document_categorization.py#L1197) and 1208 there are steps for transforming all values (image, if present, document id, labels etc) into LongTensors, so it is also for uniformity.

konfuzio_sdk/trainer/document_categorization.py
iftwigs: 1 - there is no need for padding in this step, because we already padded the text when tokenizing with BertTokenizerFast, but this step also transforms the encoded text and changes tensor's shape and format so I kept it for bert-processed text, too. I will add a todo into the refactoring ticket to check if the needed transformations can be done without padding again.
2 - i will change the value and push it

tests/trainer/test_document_categorization.py
iftwigs: since there's already a separate class for testing different bert-based models, I think it is okay to have a default 'bert-base-german-cased' here. `self.text_class.name` does not work because `self.text_class` here is not `BERT(name='smth')`, it's just `BERT` as a class, not initialized, so `BERT.name` does not exist at that point.

MohamedAmineDHIAB (approved):

docs/sdk/tutorials/document_categorization/index.md
```diff
@@ -110,20 +110,28 @@ default value will be applied.
...
+| None                   | None              | None                            | EfficientNet    | efficientnet_b3          |
+| None                   | None              | None                            | VGG             | vgg11                    |
+| None                   | None              | None                            | VGG             | vgg13                    |
+| None                   | None              | None                            | VGG             | vgg16                    |
+| None                   | None              | None                            | VGG             | vgg19                    |
+| WhitespaceTokenizer    | NBOW              | None                            | None            | None                     |
+| WhitespaceTokenizer    | NBOWSelfAttention | None                            | None            | None                     |
+| WhitespaceTokenizer    | LSTM              | None                            | None            | None                     |
+| BertTokenizerFast      | BERT              | 'bert-base-german-cased'        | None            | None                     |
```
MohamedAmineDHIAB, line 130: Can we check if we can generalize more on this table ? (maybe include links to transformers classes mentioned here:
 https://huggingface.co/docs/transformers/v4.35.2/en/model_doc/auto#transformers.AutoTokenizer

MohamedAmineDHIAB (dismissed):

zypriafl (approved):

PR #436, opened 2023-11-28
Reworked Categorization AI Training Logs

MohamedAmineDHIAB:
- [x] Removed: logs showing wrong results for test accuracy

```diff 
- logger.info(f'test_loss: {np.mean(train_losses):.3f}, test_acc: {np.nanmean(train_loss):.3f}')
```

- [x] Added: computation and training logs for `f1_score` and `accuracy_score`

```diff 
+            f1_score = f1_metric.compute(predictions=predictions_list, references=ground_truth_list, average="macro")[
+                "f1"
+            ]  # compute f1 score
+            acc_score = acc_metric.compute(predictions=predictions_list, references=ground_truth_list)[
+                "accuracy"
+            ]  # compute accuracy score
```

```diff 
+ logger.info(
+                f'Epoch: {epoch + 1} | '
+                f'Train Loss: {np.mean(train_loss):.3f} | '
+                f'Train F1: {f1_score:.3f} | '
+                f'Train Accuracy: {acc_score:.3f} |'
+          )
```

- [x] Reworked: some training logs (logging number of `Training Documents` and `Training Pages` instead of just logging `Training Examples` which is unintuitive)

### DIFF
--- a/konfuzio_sdk/trainer/document_categorization.py
+++ b/konfuzio_sdk/trainer/document_categorization.py
@@ -25,6 +25,7 @@ from konfuzio_sdk.tokenizer.base import AbstractTokenizer
 from konfuzio_sdk.tokenizer.regex import WhitespaceTokenizer
 from konfuzio_sdk.data import Document, Page, Category, CategoryAnnotation
 from konfuzio_sdk.extras import (
+    evaluate,
     timm,
     torch,
     torchvision,
@@ -909,7 +910,7 @@ class PageMultimodalCategorizationModel(PageCategorizationModel):
 
 def get_optimizer(classifier: PageCategorizationModel, config: dict) -> Optimizer:
     """Get an optimizer for a given Model given a config."""
-    logger.info('getting optimizer')
+    logger.info('Getting optimizer')
 
     # name of the optimizer, i.e. SGD, Adam, RMSprop
     optimizer_name = config['name']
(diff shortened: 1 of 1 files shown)


konfuzio_sdk/trainer/document_categorization.py
```diff
@@ -1286,14 +1293,28 @@ def _fit_classifier(
...
+        acc_metric = evaluate.load("accuracy")
+        logger.info('Begin fitting')
         best_valid_loss = float('inf')
         train_loss = float('inf')
         for epoch in range(n_epochs):
-            train_loss = self._train(train_examples, loss_fn, optimizer)  # train epoch
+            train_loss, predictions_list, ground_truth_list = self._train(
+                train_examples, loss_fn, optimizer
+            )  # train epoch
+            f1_score = f1_metric.compute(predictions=predictions_list, references=ground_truth_list, average="macro")[
```
iftwigs, line 1305: is it possible to use `konfuzio_sdk.evaluate.CategorizationEvaluation` here and in accuracy calculation instead?

konfuzio_sdk/trainer/document_categorization.py
```diff
@@ -1311,12 +1332,10 @@ def _fit_classifier(
             if patience_counter > patience:
                 logger.info('lost patience!')
                 break
-        logger.info('training finished, begin testing')
+        logger.info('Training finished. Loading best model')
```
iftwigs, line 1335: Loading the best model (sorry for nitpicking)

zypriafl (approved):

konfuzio_sdk/trainer/document_categorization.py
MohamedAmineDHIAB: Yes, I was thinking of that in the beginning but since `train_examples` and `predictions` are on the pages level and not the documents I decided to directly compute the f1 score with `evalute` (while `CategorizationEvaluation` requires the documents as a SDK Document class)

PR #455, opened 2024-01-08
Fixed: Normalization for pseudo-float values

iftwigs:
Values that are wrongly recognized as float such as `165,,` are normalized to `None` and do not cause errors when processed during Information Extraction pipeline.


konfuzio_sdk/normalize.py
```diff
@@ -231,6 +238,13 @@ def _normalize_string_to_absolute_float(offset_string: str) -> Optional[float]:
     elif offset_string.count('.') == 1 and (len(offset_string) - offset_string.index('.')) == 3:
         if all(x.isdecimal() for x in offset_string.split('.')):
             _float = float(offset_string)  # => 12.34
+    elif (
```
zypriafl, line 241: Please add a comment to explain what we want to achieve with this.

zypriafl (approved):

PR #466, opened 2024-02-27
Added: Annotation creation using Bbox only

iftwigs:
Added the possibility to create Annotations by supplying only Bbox coordinates instead of mandatory supplying Spans.


konfuzio_sdk/data.py
```diff
@@ -2280,8 +2288,37 @@ def __init__(
...
+                    'top': bbox.page.height - bbox.y1,
+                    'bottom': bbox.page.height - bbox.y0,
+                    'x0': bbox.x0,
+                    'x1': bbox.x1,
+                    'y0': bbox.y0,
+                    'y1': bbox.y1,
+                }
+                for bbox in self._bboxes
+            ]
+            merged = merge_bboxes(bboxes=bbox_dicts)
```
iftwigs, line 2307: not necessary to merge, keep spans separate

konfuzio_sdk/data.py
```diff
@@ -2283,14 +2287,21 @@ def __init__(
...
         # TODO START LEGACY to support multiline Annotations
         bboxes = kwargs.get('bboxes', None)
         if bboxes and len(bboxes) > 0:
             for bbox in bboxes:
+                # todo add comment about why we have _bboxes and why we have the checks below
                 if 'start_offset' in bbox.keys() and 'end_offset' in bbox.keys():
                     Span(start_offset=bbox['start_offset'], end_offset=bbox['end_offset'], annotation=self)
+                elif 'x0' in bbox.keys() and 'x1' in bbox.keys() and 'y0' in bbox.keys() and 'y1' in bbox.keys():
+                    page = self.document.get_page_by_index(bbox['page_index'])
+                    _ = Bbox(page=page, x0=bbox['x0'], x1=bbox['x1'], y0=bbox['y0'], y1=bbox['y1'])
```
iftwigs, line 2301: replace with the name

konfuzio_sdk/data.py
```diff
@@ -2301,8 +2312,57 @@ def __init__(
...
+                    'top': bbox.page.height - bbox.y1,
+                    'bottom': bbox.page.height - bbox.y0,
+                    'x0': bbox.x0,
+                    'x1': bbox.x1,
+                    'y0': bbox.y0,
+                    'y1': bbox.y1,
+                }
+                for bbox in self._bboxes
+            ]
+            document_bbox_dicts = {
```
iftwigs, line 2329: move it to the document class

konfuzio_sdk/data.py
```diff
@@ -2283,14 +2287,21 @@ def __init__(
...
         bboxes = kwargs.get('bboxes', None)
         if bboxes and len(bboxes) > 0:
             for bbox in bboxes:
+                # todo add comment about why we have _bboxes and why we have the checks below
                 if 'start_offset' in bbox.keys() and 'end_offset' in bbox.keys():
                     Span(start_offset=bbox['start_offset'], end_offset=bbox['end_offset'], annotation=self)
+                elif 'x0' in bbox.keys() and 'x1' in bbox.keys() and 'y0' in bbox.keys() and 'y1' in bbox.keys():
+                    page = self.document.get_page_by_index(bbox['page_index'])
+                    _ = Bbox(page=page, x0=bbox['x0'], x1=bbox['x1'], y0=bbox['y0'], y1=bbox['y1'])
+                    self._bboxes.append(_)
```
iftwigs, line 2302: replace this with the logic below so that _bboxes can be deleted

konfuzio_sdk/data.py
```diff
@@ -2602,15 +2668,13 @@ def delete(self, delete_online: bool = True) -> None:
...
-                page=self.page,
-            )
-        return self._bbox
+        self._bbox = Bbox(
+            x0=min([span.bbox().x0 for span in self.spans]),
+            x1=max([span.bbox().x1 for span in self.spans]),
+            y0=min([span.bbox().y0 for span in self.spans]),
+            y1=max([span.bbox().y1 for span in self.spans]),
+            page=self.page,
+        )
```
iftwigs, line 2677: add return

konfuzio_sdk/data.py
```diff
@@ -2287,19 +2287,50 @@ def __init__(
...
                 if 'start_offset' in bbox.keys() and 'end_offset' in bbox.keys():
                     Span(start_offset=bbox['start_offset'], end_offset=bbox['end_offset'], annotation=self)
                 elif 'x0' in bbox.keys() and 'x1' in bbox.keys() and 'y0' in bbox.keys() and 'y1' in bbox.keys():
                     page = self.document.get_page_by_index(bbox['page_index'])
-                    _ = Bbox(page=page, x0=bbox['x0'], x1=bbox['x1'], y0=bbox['y0'], y1=bbox['y1'])
-                    self._bboxes.append(_)
+                    cur_bbox = Bbox(page=page, x0=bbox['x0'], x1=bbox['x1'], y0=bbox['y0'], y1=bbox['y1'])
+                    if not self._document_bboxes:
+                        self._document_bboxes = self.document.bbox_dict
+                    bbox_dict = {
```
iftwigs, line 2304: check if it should be a property of Bbox class / replaced by an existing property

konfuzio_sdk/data.py
```diff
@@ -2287,19 +2287,50 @@ def __init__(
...
         if bboxes and len(bboxes) > 0:
             for bbox in bboxes:
-                # todo add comment about why we have _bboxes and why we have the checks below
                 if 'start_offset' in bbox.keys() and 'end_offset' in bbox.keys():
                     Span(start_offset=bbox['start_offset'], end_offset=bbox['end_offset'], annotation=self)
                 elif 'x0' in bbox.keys() and 'x1' in bbox.keys() and 'y0' in bbox.keys() and 'y1' in bbox.keys():
                     page = self.document.get_page_by_index(bbox['page_index'])
-                    _ = Bbox(page=page, x0=bbox['x0'], x1=bbox['x1'], y0=bbox['y0'], y1=bbox['y1'])
-                    self._bboxes.append(_)
+                    cur_bbox = Bbox(page=page, x0=bbox['x0'], x1=bbox['x1'], y0=bbox['y0'], y1=bbox['y1'])
```
iftwigs, line 2301: make it into a private method

iftwigs: cur_bbox -> bbox

konfuzio_sdk/data.py
```diff
@@ -2287,19 +2287,50 @@ def __init__(
...
+                    if not self._document_bboxes:
+                        self._document_bboxes = self.document.bbox_dict
+                    bbox_dict = {
+                        'page_index': cur_bbox.page.index,
+                        'top': cur_bbox.page.height - cur_bbox.y1,
```
iftwigs, line 2306: maybe reuse bbox_dict and add missing fields like page_index, top_ bottom

konfuzio_sdk/data.py
iftwigs: remove cur_bbox?

konfuzio_sdk/data.py
```diff
@@ -2287,19 +2287,50 @@ def __init__(
...
+                        'bottom': cur_bbox.page.height - cur_bbox.y0,
+                        'x0': cur_bbox.x0,
+                        'x1': cur_bbox.x1,
+                        'y0': cur_bbox.y0,
+                        'y1': cur_bbox.y1,
+                    }
+                    # checking that Bboxes provided as input contain text inside
+                    merged = get_merged_bboxes(
```
iftwigs, line 2314: merged -> merged_bboxes

konfuzio_sdk/data.py
```diff
@@ -2287,19 +2287,50 @@ def __init__(
...
+                        doc_bbox=self._document_bboxes, bboxes=[bbox_dict], doc_text=self.document.text
+                    )
+                    for merged_bbox in merged:
+                        _ = Bbox(
```
iftwigs, line 2318: move to try execpt

konfuzio_sdk/data.py
iftwigs: get_spans_from_bbox

konfuzio_sdk/data.py
```diff
@@ -2283,16 +2300,15 @@ def __init__(
...
-        bboxes = kwargs.get('bboxes', None)
-        if bboxes and len(bboxes) > 0:
-            for bbox in bboxes:
-                if 'start_offset' in bbox.keys() and 'end_offset' in bbox.keys():
-                    Span(start_offset=bbox['start_offset'], end_offset=bbox['end_offset'], annotation=self)
-                else:
-                    raise ValueError(f'SDK cannot read bbox of Annotation {self.id_} in {self.document}: {bbox}')
+        input_bbox_dicts = kwargs.get('bboxes', None)
+        if input_bbox_dicts and len(input_bbox_dicts) > 0:
+            self._add_bbox_annotation(input_bbox_dicts=input_bbox_dicts)
```
MohamedAmineDHIAB, line 2309: I would call it _add_annotation_from_bbox_dict()
It's length but it's much more explanatory than _add_bbox_annotation() which would raise the question "what is a Bbox Annotation ?"

konfuzio_sdk/data.py
```diff
@@ -2425,6 +2441,55 @@ def label_set(self) -> LabelSet:
         """Return Label Set of Annotation."""
         return self.annotation_set.label_set
 
+    def _add_bbox_annotation(self, input_bbox_dicts: dict) -> None:
+        """
+        Check if input Bboxes are suitable to create an Annotation and if they are, create it.
```
MohamedAmineDHIAB, line 2446: Since this was moved to a separate private method (which is much clearer and easier to maintain) I would use the chance to explain what it's doing in the doc string). again using "Bboxes" with upper "B" in "input Bboxes" would raise the question "what is an input Bbox" is there "a non-input Bbox" ? it can be confused with the class Bbox.

I would suggest this docstring:

<< Based on a list of dictionaries where each contains metadata for a bbox, it adds Annotations to the Document. If the dictionary has the fields "start_offset" and "end_offset" the method creates a Span based on such fields. If not, it parses the bbox coordinates from the dictionary keys and based on them, looks for Spans that could be contained within such bbox. If no Span was found, a KeyError is raised. >>

MohamedAmineDHIAB (approved):

PR #486, opened 2024-04-25
Added: functionality to tune hyperparameters for Label Set classifier independently

MohamedAmineDHIAB:
* [x] Label Set Classifier now have independent hyperparameters that can be passed as keyword arguments when instantiating the `RFExtractionAI()` by the server (which can make them parameterizable from the UI)
* [x] Label Set Classifier hyperparameters are logged just like the rest of `RFExtractionAI()` parameters

### DIFF
--- a/konfuzio_sdk/trainer/information_extraction.py
+++ b/konfuzio_sdk/trainer/information_extraction.py
@@ -1415,7 +1415,13 @@ class GroupAnnotationSets:
         if not res_dict:
             logger.warning('res_dict is empty')
             return res_dict
-        n_nearest = self.label_set_n_nearest_template
+        n_nearest = (
+            self.label_set_n_nearest_template
+            if hasattr(self, 'label_set_n_nearest_template')
+            else self.n_nearest_template
+            if hasattr(self, 'n_nearest_template')
+            else 5
+        )
         feature_df = self.build_document_template_feature_X(text, self.dict_to_dataframe(res_dict)).filter(
             self.template_feature_list, axis=1
         )
--- a/konfuzio_sdk/trainer/information_extraction.py
+++ b/konfuzio_sdk/trainer/information_extraction.py
@@ -1418,9 +1418,7 @@ class GroupAnnotationSets:
         n_nearest = (
             self.label_set_n_nearest_template
             if hasattr(self, 'label_set_n_nearest_template')
-            else self.n_nearest_template
-            if hasattr(self, 'n_nearest_template')
-            else 5
+            else (self.n_nearest_template if hasattr(self, 'n_nearest_template') else 0)
         )
         feature_df = self.build_document_template_feature_X(text, self.dict_to_dataframe(res_dict)).filter(
             self.template_feature_list, axis=1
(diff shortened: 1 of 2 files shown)


konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1158,9 +1158,9 @@ class GroupAnnotationSets:
 
     def __init__(self):
         """Initialize TemplateClf."""
```
iftwigs, line 1160: maybe we can get rid of the obsolete naming like Template, Section? in this docstring and in the class in general

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1158,9 +1158,9 @@ class GroupAnnotationSets:
...
-        self.n_nearest_template = 5
-        self.max_depth = 100
-        self.n_estimators = 100
+        self.label_set_max_depth = 100
+        self.label_set_n_estimators = 100
+        self.label_set_n_nearest_template = 5
         self.label_set_clf = None
```
iftwigs, line 1164: can we have the documentation for each of these hyperparameters either here or in dev section on Information Extraction?

konfuzio_sdk/trainer/information_extraction.py
```diff
@@ -1570,14 +1573,22 @@ def __init__(
...
-        self.first_word = first_word
         self.max_depth = max_depth
         self.n_estimators = n_estimators
+        self.use_separate_labels = use_separate_labels
         self.no_label_limit = no_label_limit
+        self.first_word = first_word
         self.n_nearest_across_lines = n_nearest_across_lines
 
+        # label set clf hyperparameters
+        self.label_set_n_nearest_template = kwargs.get('label_set_n_nearest_template', 5)
```
iftwigs, line 1585: does it make sense to move this block above the line 

`GroupAnnotationSets.__init__(self)`

and pass these values directly as an input to the class?

MohamedAmineDHIAB: @iftwigs changing class inputs or modifying attributes of a class would break backwards compatibility. It would require extra boilerplate to handle two cases so no unfortunately.

konfuzio_sdk/trainer/information_extraction.py
MohamedAmineDHIAB: The documentation will be on help together with the rest of the explanation of the other Extraction AI parameters. (cf. gitlab [ticket](https://git.konfuzio.com/konfuzio/objectives/-/issues/12490))

konfuzio_sdk/trainer/information_extraction.py
MohamedAmineDHIAB: It's out of scope for this PR

MohamedAmineDHIAB: Also bear in mind the last comment down :point_down: (Any change to class attributes / variables is very sensitive and would break the functioning saved pickles, it would also require extra boilerplate to handle old+new cases)

MohamedAmineDHIAB:
@iftwigs I like your title suggestion more so I have changed the PR's title accordingly

konfuzio_sdk/trainer/information_extraction.py
iftwigs: okay, valid point about the extra code required.

iftwigs (approved):

iftwigs (approved):

PR #487, opened 2024-04-29
Fixed: slashes are filtered out of Category names/AI saving paths

iftwigs:
Slashes are now filtered out of Category names in order to not be included into AI saving paths - it was breaking Extraction AI training pipeline before.


nengelmann (approved):
Just added a minor suggestion.
Tested and works just fine! 👍

nengelmann (approved):

PR #502, opened 2024-06-06
Added: optimized confidence thresholds for Labels

iftwigs:
A step in Extraction AI evaluation that allows to calculate an optimal threshold for a Label to achieve the best possible F1-score/precision/recall.


konfuzio_sdk/evaluate.py
```diff
@@ -82,8 +82,15 @@ def grouped(group, target: str):
...
 def compare(
-    doc_a, doc_b, only_use_correct=False, use_view_annotations=False, ignore_below_threshold=False, strict=True
-) -> pandas.DataFrame:
+    doc_a,
+    doc_b,
+    only_use_correct=False,
+    use_view_annotations=False,
+    ignore_below_threshold=False,
+    strict=True,
+    use_custom_threshold=False,
```
iftwigs, line 91: remove double arguments, leave only custom_threshold

konfuzio_sdk/evaluate.py
```diff
@@ -414,10 +435,70 @@ def calculate(self):
                 strict=self.strict,
                 use_view_annotations=self.use_view_annotations,
                 ignore_below_threshold=self.ignore_below_threshold,
+                use_custom_threshold=False,
```
iftwigs, line 438: remove because by default it's false anyway

PR #506, opened 2024-06-17
Fixed: wrong model name in categorization documentation

iftwigs:


MohamedAmineDHIAB (commented):
For `bert-base-german-cased` do we also need to change it ?

PR #510, opened 2024-07-01
Update: Colab Demo

nengelmann:
- Fix issue with special characters in passwords.
- Improve initialization check.
- Update example use case to enhance the user intro into the SDK.


notebooks/Get_started_with_the_Konfuzio_SDK.ipynb
```diff
@@ -26,319 +28,312 @@
...
-      "outputs": [],
+      "execution_count": null,
+      "outputs": []
+    },
+    {
+      "cell_type": "markdown",
       "source": [
-        "from konfuzio_sdk.data import Project"
-      ]
+        "### Uplaod and sync a new Document to Konfuzio Server"
```
iftwigs, line 226: *Upload

iftwigs: also what do you mean by sync here and in the following print statement? if you refer to the argument in `Document.from_file`, it just returns the document only after it has been fully processed (as opposed to `sync=False` which returns the document immediately, not waiting for the processing).

notebooks/Get_started_with_the_Konfuzio_SDK.ipynb
```diff
@@ -26,319 +28,312 @@
...
+      "execution_count": null,
+      "outputs": []
+    },
+    {
+      "cell_type": "markdown",
       "source": [
-        "# get project\n",
-        "project = Project(id_=project_id)"
-      ]
+        "### Get and show character level bounding boxes of the document\n",
```
iftwigs, line 271: maybe this could be reworded in a less technical way? we all know what are character-level bboxes but a new user might not... perhaps something like "Get and show OCR-recognized characters"?

iftwigs: and also maybe a small block of text before the code to explain that the document undergoes OCR when created and here is the result of it etc etc

notebooks/Get_started_with_the_Konfuzio_SDK.ipynb
```diff
@@ -26,319 +28,312 @@
...
-        "for label in labels:\n",
-        "  print(label)"
-      ]
+        "# **Documentation and Tutorials** 📄\n",
+        "\n",
+        "Konfuzio SDK comes with an extensive documentation and accompanying tutorials. \\\n",
+        "Check it out here 📣\n",
+        "\n",
+        "- [Development Documentation](https://dev.konfuzio.com/index.html)\n",
+        "- [Python API Refernce](https://dev.konfuzio.com/sdk/sourcecode.html#api-reference)\n",
```
iftwigs, line 311: Reference*

nengelmann:
@iftwigs thanks for the feedback, is now included!

iftwigs (approved):

PR #534, opened 2024-08-20
Revert "Fixed: inconsistent evaluation of Spans containing more than one word"

MohamedAmineDHIAB:
For the moment, and from a strategic perspective, there was no legitimate reason to make a big change to the Extraction AI Evaluation, instead we could just log a warning about the inconsistent evaluation results for Custom Extraction AIs that are not applying the right post-processing to their predictions. 

* [x] Added WARNING message once trying to create an `Annotation` using two or more `Spans` that could be merged into one `Span` (for example separated by tabs or whitespaces)
* [x] Increased the logging level for the integrity of SDK to WARNING instead of INFO to avoid having very lengthy evaluation logs (currently above 2000 lines of logs for a sample evaluation cf. [12588](https://git.konfuzio.com/konfuzio/objectives/-/issues/12588))
* [x] Changed WARNING about 'TP and FP are zero, precision is set to 0.' to an INFO
* [x] Check that servers logging level is not affected
* [x] Check server tests pass successfully

### DIFF
--- a/konfuzio_sdk/utils.py
+++ b/konfuzio_sdk/utils.py
@@ -341,6 +341,11 @@ def slugify(value):
     return re.sub(r'[-\s\:\.]+', '-', value).replace('-_', '_')
 
 
+def is_composed_of_spaces_and_tabs(s: str) -> bool:
+    """Check if the string contains only spaces and tabs using a regex"""
+    return re.fullmatch(r'[ \t]*', s) is not None
+
+
 def normalize_name(value: str) -> str:
     """
     Normalize names for different Konfuzio concepts by removing slashes and checking for non-ascii symbols.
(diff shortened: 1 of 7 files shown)


konfuzio_sdk/utils.py
```diff
@@ -341,6 +341,11 @@ def slugify(value):
     return re.sub(r'[-\s\:\.]+', '-', value).replace('-_', '_')
 
 
+def is_composed_of_spaces_and_tabs(s: str) -> bool:
```
iftwigs, line 344: do we want to also check if there are `\n` instances in it? i know we want to use it specifically for checking spans now, and spans can't have newlines inside, but since we also place this function into utils, i assume it should be generic.

MohamedAmineDHIAB: I agree that could be interesting, but let's keep it like this until we have a use case or necessity for the change. Once we have one we can update this utility method. Just keep an eye on it.

PR #536, opened 2024-08-22
Fixed: limit pkl_name to 63 characters

mexicat:
- [x] Bento tag names > 63 characters are not supported, so we limit the auto-generated `pkl_name` to take this into consideration.


nengelmann (approved):
Looks good and works.

zypriafl (approved):

